PR #8, opened 2023-05-08
Added requirements.txt that was missing from original repo

AugustusCeasar:
the requirements.txt from the original repo was missing because it got caught up in default .gitignore vals. Granted it an exception and added back in


ccstan99 (approved):
Thanks!

Issue #10, opened 2023-05-17
Pull Stampy info directly from coda


ccstan99:
Sample code on how to pull updated Stampy data from data:
https://github.com/StampyAI/stampy-nlp/blob/main/src/stampy_nlp/utilities/coda_utils.py

Thomas-Lemoine:
Just to be sure, is this a good time to close the issue

ccstan99:
Yep! A little trick is in your PR or commits, you can mention this issue by number to associate them.

PR #12, opened 2023-05-28
makes stampy updatable again using coda.io

Thomas-Lemoine:


align_data/stampy/__init__.py
```diff
@@ -2,5 +2,7 @@
 
 STAMPY_REGISTRY = [
     Stampy(name = 'stampy',
-        index_url = "https://stampy.ai/wiki/Special:Ask/mainlabel%3D/format%3Djson/searchlabel%3DJSON/template%3DACard/userparam%3DQA/sort%3DStampCount/order%3Ddesc/offset%3D0/limit%3D500/-5B-5BCategory:Answers-5D-5D-20-5B-5BOutOfScope::false-5D-5D-20-5B-5BCanonical::true-5D-5D/-3F%3D-2D-23-2D/-3FAnswer/-3FStampedBy/-3FTags-23-2D/prettyprint%3Dtrue/unescape%3Dtrue"),
+        coda_doc_id = 'fau7sl2hmG',
+        on_site_table = "table-aOTSHIz_mN"
+        ),
```
ccstan99, line 7: @mruwnik What's the difference between this table and 'table-YvPEyAXl8a'? Is one better than the other?

requirements.txt
```diff
@@ -19,4 +19,5 @@ python-docx
 transformers
 openpyxl
 seaborn
-matplotlib
\ No newline at end of file
+matplotlib
+codaio
```
ccstan99, line 23: @mruwnik Are there pros & cons to using codaio vs requests.get from url?

ccstan99 (commented):
Thanks so much Thomas! Looks good, but I'm not set up to run and test this, so I'll defer to @mruwnik.

align_data/stampy/__init__.py
Thomas-Lemoine: In the stampy ui code I found this:
QUESTION_DETAILS_TABLE = 'grid-sync-1059-File'
INITIAL_QUESTIONS_TABLE = 'table-yWog6qRzV4'
ON_SITE_TABLE = 'table-aOTSHIz_mN'
ALL_ANSWERS_TABLE = 'table-YvPEyAXl8a'
INCOMING_QUESTIONS_TABLE = 'grid-S_6SYj6Tjm'

and the on-site-table contains only the entries that end up on the stampy.ai site, so they're up-to-date including the answer, in the "Rich Text" key. Whereas all-answers-table, which is table-YvPEyAXl8a, has entries like:
{'Edit Answer': "Front page : I'd like to know a bunch of the technical details", 'Status': 'Not started', 'Thumbnail': 'image.jpeg', 'Tags': '', 'Related Answers': '', 'All Phrasings': "Front page : I'd like to know a bunch of the technical details\n", 'UI ID': '9IDQ', 'Link': 'https://docs.google.com/document/d/1Amcrdyi08J8yp_qoyUC9FdEMh-y1FllUlP7aGtMRqG4/edit?usp=drivesdk', 'Column 9': ''}
which doesn't have a "Rich Text" section, aka no answer to the question. So it seemed to me that the on-site-table was better for the dataset but it's also easy to change to another of the 54 other tables.

README.md
```diff
@@ -105,6 +105,13 @@ As we said in the previous section, all entries have the `text` key which contai
 
 2. `alignment_text`: This is label specific to the arXiv papers. We added papers to the dataset using Allen AI's SPECTER model and included all the papers that got a confidence score of over 75%. However, since we could not verify with certainty that those papers where about alignment, we've decided to create the `alignment_text` key with the value `"pos"` when we manually labeled it as an alignment text and `"unlabeled"` when we have not labeled it yet. Additionally, we've only included the `text` for the `"pos"` entries, not the `"unlabeled"` entries.
 
+## Running the code
+
+To update the stampy portion of the dataset, you will need a Coda token. go to coda.io, log in, and generate an API token in your account settings. Add restrictions: Doc or table, Read only, for the doc with url https://coda.io/d/_dfau7sl2hmG. Then, create a .env file at the root of the alignment research dataset, and write CODA_TOKEN="<coda_token>". It will be accessible in align_data/stampy/stampy.py
+
+
+When wishing to update the whole dataset, run `python3 main.py fetch_all`. You can also fetch a specific subsection of a dataset by its name, for example `python3 main.py fetch --dataset_name stampy`
```
mruwnik, line 113: [NIT] I'd suggest changing `python3` to `python` - anyone running this should already have setup the environment properly. Though this is mainly a matter of style, so totally up to you

align_data/stampy/__init__.py
```diff
@@ -2,5 +2,7 @@
 
 STAMPY_REGISTRY = [
     Stampy(name = 'stampy',
-        index_url = "https://stampy.ai/wiki/Special:Ask/mainlabel%3D/format%3Djson/searchlabel%3DJSON/template%3DACard/userparam%3DQA/sort%3DStampCount/order%3Ddesc/offset%3D0/limit%3D500/-5B-5BCategory:Answers-5D-5D-20-5B-5BOutOfScope::false-5D-5D-20-5B-5BCanonical::true-5D-5D/-3F%3D-2D-23-2D/-3FAnswer/-3FStampedBy/-3FTags-23-2D/prettyprint%3Dtrue/unescape%3Dtrue"),
+        coda_doc_id = 'fau7sl2hmG',
+        on_site_table = "table-aOTSHIz_mN"
```
mruwnik, line 6: how about having these values read in from env variables, but with these as defaults? I'd also suggest having a dedicated `stampy/settings.py` module that contains them all, which would pretty much consist only of things like:
```
CODA_DOC_ID = os.environ.get("CODA_DOC_ID", "fau7sl2hmG")
ON_SITE_TABLE = os.environ.get('CODA_ON_SITE_TABLE', 'table-aOTSHIz_mN')
``` 
Or even just 
```
CODA_DOC_ID = "fau7sl2hmG"
ON_SITE_TABLE = 'table-aOTSHIz_mN'
```

align_data/stampy/stampy.py
```diff
@@ -4,48 +4,64 @@
 from align_data.common.alignment_dataset import AlignmentDataset , DataEntry
 from tqdm import tqdm
 
+import os
+from dotenv import load_dotenv
```
mruwnik, line 8: nice!

align_data/stampy/stampy.py
```diff
@@ -4,48 +4,64 @@
...
 
     def fetch_entries(self):
         self.setup()
-        entries = dict(requests.get(self.index_url).json())
-        for ii, entry in enumerate(tqdm(entries["results"].keys())):
-            if self._entry_done(entry):
+
+
+        coda = Coda(CODA_TOKEN)
+        doc = Document(self.coda_doc_id, coda=coda)
```
mruwnik, line 30: shouldn't these go into the `setup` method?

align_data/stampy/stampy.py
```diff
@@ -4,48 +4,64 @@
...
+        table = doc.get_table(self.on_site_table)
+        entries = table.to_dict() # a list of dicts
+        #
+        # keys for each element of the list of dicts: 
+        # {'Edit Answer', 'Status', 'aisafety.info Link', 'External Source', 'Tags', 'Related IDs', 'Alternate Phrasings', 'Rich Text', 'Doc Last Edited', 'Question', 'Link', 'UI ID', 'Related Answers'}
+        for ii, entry in enumerate(tqdm(entries)):
```
mruwnik, line 36: why enumerate? `tqdm` should handle the process bar thingy, and logging e.g. the pageid rather than number would provide more info in the case of problems

align_data/stampy/stampy.py
```diff
@@ -4,48 +4,64 @@
...
-            qa_entry = entries["results"][entry]
-            qa_entry["question"] = ' '.join(entry.split("to ")[1:])
-            qa_entry["answer"] = entries["results"][entry]["printouts"]["Answer"]
-            qa_entry["text"] = "Question: " + qa_entry["question"] + "\n\nAnswer: " + entries["results"][entry]["printouts"]["Answer"][0]
-            # if there is more than one answer, add the rest
-            for jj in range(1, len(entries["results"][entry]["printouts"]["Answer"])):
-                qa_entry["text"] += f"\n\nAnswer {str(jj)}: " + entries["results"][entry]["printouts"]["Answer"][jj]
 
+            question = entry.get('Question', '')
+            answer = entry.get('Rich Text', '')  # Assuming 'Answer' is present in each dict
```
mruwnik, line 42: it should be. If not then either the Coda table has been changed (e.g. column hidden or renamed. Yes, this happens...) or the entry should be skipped. If it's safe to assume that this is always present, then its absence should be either logged as an error, or an exception raised. If it's not safe to assume it's always present (e.g. all answers are processed, even those which haven't been started) then you should filter them out along with the already processed questions

align_data/stampy/stampy.py
```diff
@@ -4,48 +4,64 @@
...
                 continue
-            qa_entry = entries["results"][entry]
-            qa_entry["question"] = ' '.join(entry.split("to ")[1:])
-            qa_entry["answer"] = entries["results"][entry]["printouts"]["Answer"]
-            qa_entry["text"] = "Question: " + qa_entry["question"] + "\n\nAnswer: " + entries["results"][entry]["printouts"]["Answer"][0]
-            # if there is more than one answer, add the rest
-            for jj in range(1, len(entries["results"][entry]["printouts"]["Answer"])):
-                qa_entry["text"] += f"\n\nAnswer {str(jj)}: " + entries["results"][entry]["printouts"]["Answer"][jj]
 
+            question = entry.get('Question', '')
```
mruwnik, line 41: this should raise an error or be skipped if missing. A question without a question is pointless

requirements.txt
mruwnik: Yes, of course. But not really :D The codaio library is pretty much just a wrapper around requests, so this should be fine

mruwnik:
The underlying code changed somewhat, hence the conflicts. [This commit](https://github.com/StampyAI/alignment-research-dataset/commit/bc1780211b61868059f5ff6acf6d446f04484386) resolves the conflicts

align_data/stampy/stampy.py
```diff
@@ -12,56 +17,41 @@
...
-
-            yield new_entry
+        return table.to_dict() # a list of dicts
+
+    def get_item_key(self, entry):
+        return entry.get('Question', '')
+
+    def process_entry(self, entry):
+        question = entry.get('Question', '')
+        answer = entry.get('Rich Text', '')  # Assuming 'Answer' is present in each dict
```
mruwnik, line 41: I'd like some error handling for this, or at least some logging, but on the other hand it seems that errors here are a bigger problem

mruwnik (approved):
Are you going to add any more changes, or can this be merged?

align_data/stampy/__init__.py
Thomas-Lemoine: I'm setting up a stampy/settings.py, am I right in thinking we would then have 
```
import os
from dotenv import load_dotenv
load_dotenv()
CODA_TOKEN = os.environ.get("CODA_TOKEN", "token not found")
CODA_DOC_ID = os.environ.get("CODA_DOC_ID", "fau7sl2hmG")
ON_SITE_TABLE = os.environ.get('CODA_ON_SITE_TABLE', 'table-aOTSHIz_mN')
```
inside the settings.py file? The coda_token is different from the others bc it doesn't have a default that works though.

mruwnik: exactly. A missing token is really only a problem when someone wants to run coda - you could then add something like the following to stampy.py:
```

    def setup(self, *args, **kwargs):
        if not CODA_TOKEN:
            raise ValueError('No Coda token provided!')
        super().setup(*args, **kwargs)
```

align_data/stampy/stampy.py
Thomas-Lemoine: I'm thinking of modifying the question line to question = entry['question'] so an error appears if there's no question for the entry (which should not happen for the on site answers table). 
Although since this applies to other keys (answer, url, date_published), I can make this change to the other keys as well. Something's definitely wrong if there is no "Doc Last Edited" key for example.

mruwnik: remove data/stampy.jsonl and rerun it. The code will ignore entries that have already been processed :)

Thomas-Lemoine: Is there a reason why we wouldn't want data/stampy.jsonl and data/stampy.txt to be deleted right before entering the new data? like when we run our program which wants to create these files, they should replace whatever was there beforehand, instead of silently doing nothing

Thomas-Lemoine: but thank you, I see that it works

mruwnik: depends. In this case maybe so, but some data sources take a long time (e.g. lesswrong posts), so it's nice for it to not reprocess stuff if not needed. It might be worth discussing whether this should be controlled with a flag or something - worth raising this during the meeting

ccstan99:
I looked at the sample output generated `stampy.jsonl`. Can you please remove the keys `question` and `answer`? They are duplicated from `title` and `text` which we are using for consistency with the rest of the dataset. Otherwise, things look good. You can browse the output at https://huggingface.co/datasets/StampyAI/alignment-research-dataset/viewer/stampy/train

align_data/stampy/stampy.py
```diff
@@ -1,46 +1,45 @@
...
-        qa_entry["question"] = ' '.join(entry.split("to ")[1:])
-        qa_entry["answer"] = entries["results"][entry]["printouts"]["Answer"]
-        qa_entry["text"] = "Question: " + qa_entry["question"] + "\n\nAnswer: " + entries["results"][entry]["printouts"]["Answer"][0]
-        # if there is more than one answer, add the rest
-        for jj in range(1, len(entries["results"][entry]["printouts"]["Answer"])):
-            qa_entry["text"] += f"\n\nAnswer {str(jj)}: " + entries["results"][entry]["printouts"]["Answer"][jj]
-
+        question = entry['Question'] # raise an error if the entry has no question
+        answer = entry['Rich Text']
+        url = entry['Link']
```
ccstan99, line 30: The Llink from coda goes to a gdoc for editing the answers. We'll want to expose something like:
`url = 'https://aisafety.info?state=' + entry['UI ID']`

align_data/alignment_newsletter/alignment_newsletter.py
```diff
@@ -37,7 +37,7 @@ def process_entry(self, row):
         """
         return DataEntry({
             "url": "https://rohinshah.com/alignment-newsletter/",
-            "source": self.name,
+            "source": "alignment newsletter",
```
ccstan99, line 40: Would you please double check all the changed files unrelated to stampy? Looks like you're about to undo the commit from my last PR.

align_data/stampy/stampy.py
```diff
@@ -1,46 +1,45 @@
...
             "authors": "n/a",
-            "date_published": "n/a",
-            "text": qa_entry["text"],
-            "question": qa_entry["question"],
-            "answer": qa_entry["answer"],
-            "entry": entry
+            "date_published": date_published,
+            "text": answer,
+            "question": question,
+            "answer": answer,
```
ccstan99, line 44: No need for 'question' & 'answer' keys here. Duplicates of 'title' and 'text' from above, which are our standard keys.

align_data/alignment_newsletter/alignment_newsletter.py
Thomas-Lemoine: A file got deleted and I didn't understand how and I tried for a while to undo that particular file deletion within the commit but I didn't manage to make it work so I undid the merge commit as a whole. The thought was that once I fix this problem I can just merge it all again, but the file is still deleted even though I reverted the merge and I am very confused

align_data/stampy/stampy.py
```diff
@@ -22,12 +22,12 @@ def items_list(self):
...
 
     def get_item_key(self, entry):
-        return entry.get('Question', '')
+        return entry['Question']
 
     def process_entry(self, entry):
         question = entry['Question'] # raise an error if the entry has no question
         answer = entry['Rich Text']
-        url = entry['Link']
+        url = url = 'https://aisafety.info?state=' + entry['UI ID']
```
mruwnik, line 30: looks like you copied too much :D

align_data/stampy/stampy.py
```diff
@@ -1,46 +1,43 @@
...
 
 logger = logging.getLogger(__name__)
 
+
 @dataclass
 class Stampy(AlignmentDataset):
 
-    index_url : str
-    done_key = "entry"
+    done_key = "question"
```
mruwnik, line 14: you should probably change this to `title`, as you removed the `question` key from the dataentry

align_data/stampy/stampy.py
Thomas-Lemoine: haha true thanks

mruwnik:
looks good. Do you want me to take over handling the conflicts etc? 

For future reference, it's a lot easier if you work directly on the source repo, rather than managing your own fork. You should probably already have write access to it, and if not, it's easy to add you.

Thomas-Lemoine:
Yes, thank you. 
Yeah I'm not making that mistake again, GPT4 advised me to create the fork and work on the feature there and it was much more confusing than it needed to be. From now on my understanding is that I create a branch in StampyAI/ARD, develop the feature, keep pulling from the main throughout in case there are incoming changes so that the final merge is easy, and that should be much simpler.

mruwnik:
yup, exactly. A fork is good if you're doing a fly-by fix, or in general don't have write access. But it makes everything harder

mruwnik:
I added a couple of changes and merged it into main. I also uploaded the resulting file to Huggingface - looks good!

ccstan99:
Nice! Just looked at the HF. We'll want to add "Stampy aisafety.info" to the authors list. 

Fixes #10.

mruwnik:
the authors will be added in https://github.com/StampyAI/alignment-research-dataset/pull/52

PR #15, opened 2023-06-05
Upload data files to Huggingface

mruwnik:
Allows processing and uploading a datasource as a github [actions button](https://github.com/StampyAI/alignment-research-dataset/actions/workflows/fetch-dataset.yml).

image

Choose the datasource from the dropdown list and press `Run workflow`. If successful, the dataset on [Huggingface](https://huggingface.co/datasets/StampyAI/alignment-research-dataset/tree/main) will be replaced by its results

### DIFF
--- a/.github/workflows/fetch-dataset.yml
+++ b/.github/workflows/fetch-dataset.yml
@@ -37,7 +37,7 @@ on:
           - yudkowsky.net
 
 jobs:
-  fetch-dataset:
+  build-dataset:
     runs-on: ubuntu-latest
 
     steps:
@@ -55,8 +55,14 @@ jobs:
     - name: Fetch dataset
       run: python main.py fetch -d ${{ inputs.datasource }}
 
-    - name: Setup Huggingface client
-      run: pip install huggingface_hub
+  upload:
+    runs-on: ubuntu-latest
+    needs: build-dataset
 
-    - name: Upload file
-      run: python upload_to_huggingface.py ${{ secrets.HUGGINGFACE_TOKEN }} data/${{ inputs.datasource }}.jsonl
+    if: github.ref == 'refs/heads/master'
+    steps:
+      - name: Setup Huggingface client
+        run: pip install huggingface_hub
+
+      - name: Upload file
+        run: python upload_to_huggingface.py ${{ secrets.HUGGINGFACE_TOKEN }} data/${{ inputs.datasource }}.jsonl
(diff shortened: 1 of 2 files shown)


.github/workflows/fetch-dataset.yml
```diff
@@ -2,18 +2,11 @@ name: Fetch Alignment Research Dataset
 
 on:
   workflow_dispatch: # allow manual triggering
```
mruwnik, line 4: should this be limited to active devs or something?

ccstan99: This uploads from the main branch of the GH repo, not from anyone's local copy or the gdrive, correct? If so, my initial thought is if we're careful about code reviews and performing some automated testing before merging code, then this *should* be fine. But if you'd like to put in some authentication, that's OK with me too. We can discuss further with the group at the dev call Tuesday if you'd like.

mruwnik: yup. Though you can specify your own branch if you want to test stuff or something. My main issue is if someone that isn't an active dev comes by and triggers it. Though I'm guessing I'm just being paranoid - it looks like you have to be a collaborator to trigger actions, so should be fine

ccstan99: It looks like I need to be logged into my account to trigger it, so maybe authentication is already built-in. However, I'd prefer NOT allowing the option to specify from an separate branch if that "deploys" to the public dataset in case there are other applications expecting reliable datasource. As long as devs can generate a dataset file locally and inspect it before deploying that should be sufficient for testing. If we need to allow for further testing, then maybe have a separate non-public test dataset if/when someone specifies a non-main branch?

mruwnik: I split the pipeline into a build phase, which can be triggered from any branch, and a deploy one which will only run on main. This way people can test whether things will build while developing stuff, but will only be able to actually push to huggingface once their code gets merged into main

mruwnik:
ping

.github/workflows/fetch-dataset.yml
ccstan99: That sounds good. Since this doesn't run locally, people won't be able to see the actual outputs but should be enough to test for potential error messages.

ccstan99 (approved):
LGTM! I can add this as a separate issue but before uploading to HuggingFace, we need to add the alignment_newsletter summary entries where it's available. I'm guessing this is just for greaterwrong and arxiv but we'll need to double-check. alignment_newsletter is technically not supposed to be separate subset on HF.

mruwnik:
yup, that sounds like a separate issue

PR #16, opened 2023-06-05
Unify data processing

mruwnik:
More unification - mainly by splitting up the processing flow into smaller methods that can be overloaded by subclasses. I checked all of the various dataset classes, and they all generate sensiblish data, apart from the `OtherBlog` type (selenium doesn't work for me and I couldn't be bothered to debug it now) and `Stampy` (it's getting overhauled anyway). I ran each type at least twice, to check whether it would skip previously processed files


mruwnik:
ping

PR #20, opened 2023-06-12
Push all gdoc files to Huggingface

mruwnik:
This adds a github action that will push all jsonl files in https://drive.google.com/drive/folders/1n4i0J4CuSfNmrUkKPyTFKJU0XWYLtRF8 to huggingface


ccstan99 (approved):
LGTM!

ccstan99 (approved):
It might be "nice" to have the option for the user to select a file/dataset (like you did for build) instead of always updating everything? As a public record of which files contents actually changed?

Also snuck in a commit to remove aipulse.org since the website doesn't exist any more. #26

mruwnik:
I'll check it in a different PR - to make sure it works

PR #31, opened 2023-06-13
Add substack scraper

mruwnik:


align_data/blogs/__init__.py
```diff
@@ -38,4 +40,12 @@
         name="waitbutwhy",
         gdrive_address="https://drive.google.com/uc?id=1z0kbDd8vDsgOH6vV9z0XBnTQutXFJ6x-",
     ),
+    SubstackBlog(
+        name="import.ai",
+        url="https://importai.substack.com"
+    ),
+    SubstackBlog(
+        name="ml.safety.newsletter",
+        url="https://newsletter.mlsafety.org"
```
ccstan99, line 49: I have a **slight** preference for dataset names to be more similar to the original domain `importai.substack.com` and `newsletter.mlsafety.org`,  just to make it easier for people to identify at a glance. But maybe these are more similar to names people use in conversation? If we're not using the domains, then maybe use underscores instead? There's a already mixed variations with underscores and dots. So maybe leave this as is and I'll just write up the dataset list for plex & Rob to review naming conventions at Friday's meeting?

.github/workflows/fetch-dataset.yml
```diff
@@ -22,9 +22,11 @@ on:
           - gdrive_ebooks
           - generative.ink
           - gwern_blog
+          - import.ai
           - intelligence.org
           - jsteinhardt.wordpress.com
           - markdown.ebooks
+          - ml.safety.newsletter
```
ccstan99, line 29: Also need to update the same for .github/workflows/push-datasets.yml but I guess you need the 'Upload single file' PR (just approved) merged first.

ccstan99 (commented):
Am I reading this correctly that generating the blogs from this branch will essentially test the previous 2 PRs as well? I'm juggling a couple things for work, but I'd like to run and check the output before approving these. There's a chance I might not get to do that for the next couple of days. LMK if you need it approved sooner though.

ccstan99:
Alternatively, if you're willing generate and upload the .jsonl files for all the blogs to HuggingFace, I can just double-check output? 😬

ccstan99 (dismissed):
Did you want to update .github/workflows/push-datasets.yml before merging this? Then also update the HF_hub dataloader with the 2 new datasets?

align_data/blogs/__init__.py
ccstan99: `import.substack.ai` and `newsletter.ml.safety`? I don't know. It's neither the domain nor what people would call it in conversation. Let's revert back to what you had before and handle renaming as a separate issue after we've decided on a general convention. There's a lot of inconsistency we need to clean up anyway.

ccstan99 (approved):

Issue #36, opened 2023-06-17
Rename datasets for consistency


ccstan99:
Generally use human friendly names for datasets instead of domain names unless latter is well known. Use _underscore/snake case to separate words in names. Use . to separate words in domain names. Will need to update dataset registry, GH actions, HF dataset loader, and README.md

current | proposal
-- | --
agentmodels | agentmodels
aiimpacts.org | aiimpacts
aipulse.org | aipulse
aisafety.camp | aisafety.camp
alignment_newsletter | alignment_newsletter
arbital | arbital
arxiv_papers | arxiv
audio_transcripts | audio_transcripts
carado.moe | carado.moe
cold.takes | cold_takes
deepmind.blog | deepmind_blog
distill | distill
eaforum | eaforum
gdocs | gdocs (rename later)
gdrive_ebooks | gdrive_ebooks (rename later)
generative.ink | generative.ink
gwern_blog | gwern_blog
import.ai | import.ai
intelligence.org | miri
jsteinhardt.wordpress.com | jsteinhardt_blog
lesswrong | lesswrong
markdown.ebooks | markdown.ebooks (rename later)
ml.safety.newsletter | ml_safety_newsletter
nonarxiv_papers | nonarxiv_papers
qualiacomputing.com | qualiacomputing
reports | reports (rename later)
stampy | aisafety.info
vkrakovna.wordpress.com | vkrakovna_blog
waitbutwhy | waitbutwhy
yudkowsky.net | yudkowsky_blog

PR #37, opened 2023-06-17
Rename sources

mruwnik:
https://github.com/StampyAI/alignment-research-dataset/issues/36#

### DIFF
--- a/.github/workflows/fetch-dataset.yml
+++ b/.github/workflows/fetch-dataset.yml
@@ -11,7 +11,6 @@ on:
           - aiimpacts
           - aipulse
           - aisafety.camp
-          - alignment_newsletter
           - arbital
           - audio_transcripts
           - carado.moe
--- a/.github/workflows/fetch-dataset.yml
+++ b/.github/workflows/fetch-dataset.yml
@@ -8,32 +8,32 @@ on:
         type: choice
         options:
           - agentmodels
-          - aiimpacts.org
-          - aipulse.org
+          - aiimpacts
+          - aipulse
           - aisafety.camp
           - alignment_newsletter
           - arbital
           - audio_transcripts
           - carado.moe
-          - cold.takes
-          - deepmind.blog
+          - cold_takes
+          - deepmind_blog
           - distill
           - gdocs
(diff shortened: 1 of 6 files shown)


.github/workflows/fetch-dataset.yml
```diff
@@ -8,32 +8,31 @@ on:
         type: choice
         options:
           - agentmodels
-          - aiimpacts.org
-          - aipulse.org
+          - aiimpacts
+          - aipulse
           - aisafety.camp
-          - alignment_newsletter
```
ccstan99, line 14: Removing alignment_newsletter since it takes ~30 minutes to scrape and shouldn't change any more.

ccstan99 (approved):
I've uploaded renamed files to gdrive. After you've updated data loader, feel free to push to HF. There's a zip file with previous jsonl, just in case I goofed. We can delete it if everything looks good on HF.

mruwnik:
fixed conflicts

align_data/blogs/__init__.py
```diff
@@ -7,25 +7,25 @@
...
     ),
-    WordpressBlog(name="vkrakovna.wordpress.com", url="https://vkrakovna.wordpress.com"),
+    WordpressBlog(name="vkrakovna_blog", url="https://vkrakovna.wordpress.com"),
     WordpressBlog(
-        name="yudkowsky.net",
+        name="yudkowsky_blog",
         url="https://yudkowsky.net",
         strip=["^\s*Download as PDF\n"]
     ),
     MediumBlog(name="deepmind.blog", url="https://deepmindsafetyresearch.medium.com/", authors=["DeepMind Safety Research"]),
```
ccstan99, line 25: We're using `deepmind_blog` with an underscore. Somehow that got undone in the last commit.

ccstan99 (approved):
And thanks for all the black magic to handle authors!

.github/workflows/fetch-dataset.yml
mruwnik: not any more :D Now it takes ~5s

Issue #35, opened 2023-06-17
Add `authors` key for all entries


ccstan99:
The stampy-chat has found including the authors field helps improve retrieval. Many of our datasets already include `authors`, we should make it available on all entries. If a individual name(s) is not available, use the name of the organization: Stampy, DeepMind, authors from blogs can be infeered, etc.

PR #38, opened 2023-06-17
Add an authors field to all sources

mruwnik:
https://github.com/StampyAI/alignment-research-dataset/issues/35

### DIFF
--- a/align_data/audio_transcripts/audio_transcripts.py
+++ b/align_data/audio_transcripts/audio_transcripts.py
@@ -22,6 +22,39 @@ class AudioTranscripts(GdocDataset):
             self.files_path.mkdir(parents=True, exist_ok=True)
             self.zip_from_gdrive(path=self.raw_data_path)
 
+    @staticmethod
+    def extract_authors(text):
+        """Attempt to extract the authors from the text.
+
+        The first line tends to be the title, which tends to contain info about who's talking,
+        so do some black magic to try to guess at the names
+        """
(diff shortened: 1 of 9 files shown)


align_data/audio_transcripts/audio_transcripts.py
```diff
@@ -22,6 +22,39 @@ def setup(self):
             self.files_path.mkdir(parents=True, exist_ok=True)
             self.zip_from_gdrive(path=self.raw_data_path)
 
+    @staticmethod
```
Aprillion, line 25: is there a benefit to use a class with static methods instead of just a function inside a module?

align_data/audio_transcripts/audio_transcripts.py
```diff
@@ -22,6 +22,39 @@ def setup(self):
...
+    def extract_authors(text):
```
Aprillion, line 26: no type hints in this repo yet? presumably `extract_authors(text: str) -> str | None:` 🤔

Aprillion (approved):

align_data/audio_transcripts/audio_transcripts.py
mruwnik: types?! In Python?! Sir, you jest!!

align_data/audio_transcripts/audio_transcripts.py
mruwnik: here mainly for consistency with the other datasets. But otherwise no.

Issue #32, opened 2023-06-14
Delete generated files when rescraping


ccstan99:
For all datasets except for arxiv and greaterwrong, delete generated .jsonl and .txt files to allow for a clean scrape from scratch. Maybe offer a command line flag, something like `-rebuild`?

PR #40, opened 2023-06-17
Optionally rebuild all when fetching

mruwnik:
https://github.com/StampyAI/alignment-research-dataset/issues/32

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # AI Alignment Research Dataset
 
-The AI Alignment Research Dataset is a collection of documents related to AI Alignment and Safety from various books, research papers, and alignment related blog posts. This is a work in progress. Components are still undergoing a cleaning process to be updated more regularly. The most current version is available on [HuggingFace StampyAI/alignment-research-dataset](https://huggingface.co/datasets/StampyAI/alignment-research-dataset). This repository is the code to reproduce it. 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -56,16 +56,30 @@ cd alignment-research-dataset
...
+The command to fetch all datasets is below. Again, the optional `--rebuild` parameter allows you to scrape everything from scratch.
+
+```sh
+python main.py fetch-all --rebuild
+```
+
+To get a summary of the merged dataset, Replace [MERGED_DATASET_PATH] with the path to the merged dataset file.
+
+```sh
+python main.py count-tokens [MERGED_DATASET_PATH]
```
ccstan99, line 82: Added some extra instructions. LMK if something doesn't look right.

main.py
```diff
@@ -15,45 +15,48 @@
...
         """
         for name in align_data.ALL_DATASETS:
             print(name)
-            self.cmd_fetch(name)
+            self.fetch(name, rebuild)
 
         return None  #merge_all_files(out_dir = self.out_path)
 
-    def cmd_count_tokens(self, merged_dataset_path: str) -> None:
+    def count_tokens(self, merged_dataset_path: str) -> None:
```
ccstan99, line 59: Can we rename `count-tokens` to `summarize`?

It'd be nice to have a summary of all the different datasets and number of row/entries in each... essentially, what HF dataset viewer sometimes lists when rows actually show up. This might be a good way to double check if both are in sync. Shall I write a separate issue?

ccstan99 (approved):

main.py
mruwnik: if you could

PR #44, opened 2023-06-18
Add missing fields

mruwnik:


align_data/distill/distill.py
```diff
@@ -1,94 +1,32 @@
...
-        body = markdown_text
-
-        # pull the ol with class references out of the soup
-        references = soup.find("ol", {"class": "references"})
-        if references:
-            # for each reference li, get the the span with the class title
-            references = [
-                {"title": reference.find("span", {"class": "title"}).text}
-                for reference in references.find_all("li")
+class Distill(RSSDataset):
```
mruwnik, line 7: dunno about this. It slows things down, but fetches the contents + doesn't need to keep the raw html in repo

ccstan99: The distill.pub website will no longer be updated, so keeping contents on gdrive is an option. Either way, the raw html shouldn't be in the repo.

align_data/alignment_newsletter/alignment_newsletter.py
```diff
@@ -35,21 +33,31 @@ def process_entry(self, row):
...
+            if not v or pd.isna(v):
+                return None
+            if cast:
+                return cast(v)
+            return v
+
         return DataEntry({
-            "url": "https://rohinshah.com/alignment-newsletter/",
-            "source": self.name,
+            "url": handle_na(row.URL) or 'n/a',
```
ccstan99, line 47: Looks like `url` is the source being summarized, correct? Just be sure when creating the IDs, these should be modified so it doesn't clash with the hashed url from the source, i.e. arxiv, alignment forum, lesswrong, etc.

ccstan99 (approved):
High level it looks fine but I'm feeling a bit Machievllian. It might be easier to just review the scraped out dataset after you've pushed it to HF.

align_data/alignment_newsletter/alignment_newsletter.py
mruwnik: yes - good point

Issue #9, opened 2023-05-17
Update arbital links


ccstan99:
Make sure arbital links and information properly updated. Leverage work done by AlignmentSearch.

henri123lemoine:
@plex We need to discuss a few issues regarding processing of information, for Arbital and other sources.

In the current (local) version, there are a few issues:
- No date published or author(s) atm, and I have trouble finding this information, though might be possible to implement.
- Arbital hyperlinks, ex: "[1e4 AI Alignment]" in the text ('[{(opt) alias} {concept}]'). Another version is "[{alias}]", like [1] or [76]. Can be replaced by concept only, or kept the same, or replaced by normal-hyperlink notation link.
- Regular hyperlinks, ex: "[test hyperlink](https://www.url.com)". Can be replaced by linked concept only, or kept the same.
- Summaries: at the beginning of some pages' text, there's a summary with format "[summary: {actual summary}]...etc". Can be removed, or extracted and saved in the dataset with the 'summary' key, or (?).
- Todos: formatted as '[todo: {actual todo}]', can be anywhere in text, does not show up on website. Can be removed, kept as is, or reformatted slightly. There are 
- Tags: not currently implemented, though it could be worth implementing. The tag would describe the arbital subspace of the page (`ARBITAL_SUBSPACES = ['ai_alignment', 'math', 'rationality']`).
- The previous link issues are resolved.

More generally, we need to determine how much we want to pre-process in the dataset-making stage. Do we want to make it as convenient as possible for people to use the dataset directly, but miss some use cases (if for some reason someone wants to use the todos), or do we want to just keep as-is during dataset-creation and then do post-processing before creating the embeddings, or a mix of both depending on likelihood of usefulness?

plexish:
Replies to the bits I have takes on:
-Date published and authors is kinda retrievable from edit history from actual arbital, but not the greaterwrong scrape. Author will (almost?) always be Yudkowsky for the alignment ones.
-Normal hyperlink notation seems reasonable, if that's what most documents use? Not familiar with standards around the dataset.
- Summary: Keep, maybe in the summary key. Though maybe pre-appending them to the document is actually pretty good here?
- Drop the todos.
- Tags would be kinda nice to have, but not important.


Generally, mix of both seems reasonable to me, e.g. todos are rare enough that this is likely not worth keeping, but summaries seem good?

ccstan99:
[...]

PR #52, opened 2023-06-20
add summary + authors to arbital

mruwnik:
https://github.com/StampyAI/alignment-research-dataset/issues/9

### DIFF
--- a/align_data/arbital/arbital.py
+++ b/align_data/arbital/arbital.py
@@ -1,3 +1,4 @@
+import re
 import logging
 import requests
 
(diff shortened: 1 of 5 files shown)


align_data/arbital/arbital.py
```diff
@@ -37,23 +119,23 @@ def get_item_key(self, item):
     def process_entry(self, alias):
         try:
             page = self.get_page(alias)
+            summary, text = extract_text(page['text'])
+
+            return DataEntry({
+                'title': page['title'] if 'title' in page else 'n/a',
+                'text': text,
+                'date_published': page.get('editCreatedAt') or page.get('pageCreatedAt') or 'n/a',
+                'url': f'https://arbital.com/p/{page.get("alias") or alias}',
```
ccstan99, line 128: So `title` or `date_published` might still be 'n/a'? Should we default to the empty string or keep 'n/a' to make it more obvious?

align_data/stampy/stampy.py
```diff
@@ -44,7 +52,7 @@ def clean_text(text):
             "source_filetype": "markdown",
             "url": url,
             "title": question,
-            "authors": [],
+            "authors": ['Stampy aisafety.info'],
```
ccstan99, line 55: Ah, you read my mind!

ccstan99 (approved):
Looks good. I'll probably look more closely at the uploaded HF files when they're ready. Thanks!

align_data/arbital/arbital.py
mruwnik: hmm. That's something that would have to be decided at some point :/ Currently it looks like sort of a mixture, with some places having empty string, and others 'n/a'

ccstan99: I have a slight personal preference for empty string. Feels somehow simpler, possibly my delusion. Seems like 'n/a' is already more consistently used elsewhere? Probably no need to go back to fix things now unless there's a compelling reason otherwise. Or it is overkill to use a constant?

Issue #51, opened 2023-06-20
Remove useless keys


ccstan99:
`paged_url` doesn't seem to be relevant and can be removed. 
`source_filetype` (epub, markdown, blog, docx, text, pdf) and `converted_with` should either be used consistently across all dataset or just removed.

Issue #48, opened 2023-06-20
Keys summary & authors


ccstan99:
The keys `summary` and `authors` should be available for all entries. Add to INIT_DICT and HF DEFAULT_FEATURES. If these are unavailable have empty list rather than null or None.

PR #54, opened 2023-06-23
Tidy up

mruwnik:
https://github.com/StampyAI/alignment-research-dataset/issues/48
https://github.com/StampyAI/alignment-research-dataset/issues/51

### DIFF
--- a/align_data/common/alignment_dataset.py
+++ b/align_data/common/alignment_dataset.py
@@ -21,6 +21,8 @@ INIT_DICT = {
     "date_published": None,
     "title": None,
     "url": None,
+    "summary": lambda: [],
+    "authors": lambda: [],
 }
 
 # Used to limit the size of the text used when generating hashes.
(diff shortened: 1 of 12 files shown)


ccstan99 (approved):
I would like to see `authors` and `summary` included in INIT_DICT and initialized to the empty list. That would ensure they exist for all entries, correct?

align_data/common/alignment_dataset.py
```diff
@@ -21,6 +21,8 @@
     "date_published": None,
     "title": None,
     "url": None,
+    "summary": lambda: [],
+    "authors": lambda: [],
```
mruwnik, line 25: they are here. It's via a lambda function, as otherwise they'd all get initiased with the same empty list. Which shouldn't be a problem, but potentially someone could later update e.g. the authors list for a single entry which didn't previously have authors, resulting in all items without authors having that item. This is a known gotcha, like in:
```
def bla(items=[]):
    items.append(1)
    return items

print(bla())         ==> [1]
print(bla())         ==> [1, 1]
print(bla())         ==> [1, 1, 1]
print(bla([1, 2])) ==> [1, 2, 1]
print(bla())         ==> [1, 1, 1, 1]
```
Fun!

Issue #57, opened 2023-06-24
Handle YouTube links


ccstan99:
Somehow we seem to be missing the 450+ YouTube sources from last years scrape. We'll need "recreate" that and also be able to [pull transcripts given a YouTube link](https://github.com/jdepoix/youtube-transcript-api) and [generate whisper transcripts](https://platform.openai.com/docs/api-reference/audio). Rob's video will be handled under audio_transcripts but we'll need to figure out a general way to keep these updated when new videos are release. There's also lots of good AI alignment discussions, interviews, and other content on YouTube.

ccstan99:
There's now a [`youtube` tab in the master gsheet](https://docs.google.com/spreadsheets/d/1pgG3HzercOhf4gniaqp3tBc3uvZnHpPhXErwHcthmbI/edit#gid=1969505904) that includes all the videos from the scrape last June. Unfortunately, it only has a really long title and some authors. I was able to reconstruct some of the urls from the id in the title. We might need to find someone to look up all the urls based on titles. However, it might be more effective to just use the 1st few links on the page to scrape entire channels (i.e. Rob Mile's AI Safety Reading Group) or other helpful playlists, which might cover a large extant of the content anyway.

ccstan99:
This [LW post has a few helpful playlists](https://www.lesswrong.com/posts/tas3QZfKm4bdnAWsZ/ai-alignment-youtube-playlists).
- https://www.youtube.com/playlist?list=PLCRVRLd2RhZTpdUdEzJjo3qhmX3y3skWA 
- https://www.youtube.com/playlist?list=PLTYHZYmxohXpn5uf8JZ2OouB1PsDJAk-x (slides--heavy)

The playlists and channels should probably be moved to the indices tab.

PR #58, opened 2023-06-26
Gsheets

mruwnik:


.github/workflows/fetch-dataset.yml
```diff
@@ -26,16 +26,16 @@ on:
           - generative.ink
           - gwern_blog
           - importai
-          - miri
+          - individual_docs
```
mruwnik, line 29: any ideas for a better name for this?

align_data/gdocs/gsheets.py
```diff
@@ -0,0 +1,444 @@
...
+    def getter(url):
+        for func in funcs:
+            if res := func(url):
+                return res
+    return getter
+
+
+PARSERS = {
+    # Unhandled items that will be caught later. Though it would be good for them also to be done properly
+    'oxford.universitypressscholarship.com': none_with_error(''),
```
mruwnik, line 215: the links to this just fail

align_data/gdocs/gsheets.py
```diff
@@ -0,0 +1,444 @@
...
+
+    # Paywalled journal
+    'linkinghub.elsevier.com': none_with_error('Elsevier is a known parasite - no point in looking to them for content'),
+    'www.dl.begellhouse.com': none_with_error('This article is paywalled'),
```
mruwnik, line 219: these only have abstracts available

align_data/gdocs/gsheets.py
```diff
@@ -0,0 +1,444 @@
...
+    ),
+    'www.ssrn.com': get_pdf_from_page('.abstract-buttons a.button-link:-soup-contains("Download")'),
+    'www.vox.com': element_extractor('did.c-entry-content', remove=['c-article-footer']),
+    'www.weforum.org': element_extractor('div.wef-0'),
+    'www6.inrae.fr': element_extractor('div.ArticleContent'),
+    'www.aleph.se': element_extractor('body'),
+    'yjolt.org': get_pdf_from_page('span.file a'),
+    'yoshuabengio.org': element_extractor('div.post-content'),
+
+    # To be implemented
```
mruwnik, line 319: these I couldn't be bothered to do now. This PR is already much too large

ccstan99 (approved):
Oh, forgot to approve this after the meeting. This was my understanding based on our discussion but feel free to correct if you find a more effective method:
1. Download PDFs to gdrive/data/raw
2. Update "messy" input file maintained/suggested by humans to log any download/reading errors or downloaded target file name
3. Keep a "clean" output file that's generated from scraper with metadata, without full text
4. Have some way to keep files synced if someone wants to run scraper and take advantage of previously downloaded/scraped files.

align_data/gdocs/gsheets.py
```diff
@@ -0,0 +1,444 @@
...
+        remove=[
+            '#section-cited-by', '.Copyright', '.issue-navigation', '.ReferencedArticles',
+            '.LicenseInfo', '.ArticleIdentifierLinks', '.Banner', '.screen-reader-main-title', '.Publication'
+        ]
+    ),
+    'www.ssrn.com': get_pdf_from_page('.abstract-buttons a.button-link:-soup-contains("Download")'),
+    'www.vox.com': element_extractor('did.c-entry-content', remove=['c-article-footer']),
+    'www.weforum.org': element_extractor('div.wef-0'),
+    'www6.inrae.fr': element_extractor('div.ArticleContent'),
+    'www.aleph.se': element_extractor('body'),
```
ccstan99, line 315: Are all the `www.` required? If not, should we remove them in case someone types in a url without it and doesn't get matched up?

align_data/gdocs/gsheets.py
ccstan99: Agreed. Focus on higher priorities first. You're only missing a few rows here and there. My brain's not functioning right now but I presume there's a generic, "This domain is not yet supported" error & skip handler?

align_data/gdocs/gsheets.py
```diff
@@ -0,0 +1,444 @@
...
+        return pdf
+
+    logger.error('No handler defined for %s - please add one, or change the url (%s)', domain, url)
+
+
+# TODO: These are domains that need handlers added
+PROBLEMATICAL_DOMAINS = [
+    'www.ssrn.com',  # pdf downloads work on the page, but not via requests
+    'papers.ssrn.com',
+    'www.jstor.org',  # JSTOR pages require accepting their terms before downloading
```
ccstan99, line 347: Downloading & storing the PDFs on our gdrive should help with some of these.

align_data/gdocs/gsheets.py
```diff
@@ -0,0 +1,444 @@
...
+    'www.judiciary.senate.gov', # not yet implemented
+    'www.goodreads.com',  # not yet implemented
+    'www.repository.cam.ac.uk', # requests gets a 502 error - needs investigating
+    'www.researchgate.net',  # researchgate needs JS to render the page properly
+    'www.youtube.com',  # not yet implemented
```
ccstan99, line 352: I added a `youtube` tab on gsheets for if/when you feel up to tackling it and updated instructions on #57 .

align_data/gdocs/gsheets.py
mruwnik: no, they get matched exactly. That being said, it would be trivial to first strip any prepending `www.` and just have the basic domain, which would handle both cases

mruwnik:
This has been totally overhauled. The current process is:
* run `python main.py update_metadata 1pgG3HzercOhf4gniaqp3tBc3uvZnHpPhXErwHcthmbI special_docs.csv 1l3azVJVukGAvZPgg0GyeqiaQe8bEMZvycBJaA8cRXf4`
* this goes through the `special_docs.csv` file and for each row, checks that all metadata fields are set, that the `source_url` points to something and that something decent can be extracted from the source url
* the results of this are written to the appropriate tab of [this](https://docs.google.com/spreadsheets/d/1l3azVJVukGAvZPgg0GyeqiaQe8bEMZvycBJaA8cRXf4/edit#gid=0) to keep things nice and tidy + all pdf get saved to https://drive.google.com/drive/folders/1etWiXPRl0QqdgYzivVIj6wCv9xj5VYoN

.github/workflows/push-datasets.yml
```diff
@@ -24,21 +24,24 @@ on:
...
           - eaforum
+          - ebooks
           - gdocs
           - gdrive_ebooks
           - generative.ink
           - gwern_blog
+          - html_articles
           - importai
-          - miri
+          - individual_docs
```
ccstan99, line 34: How is individual_docs different from ebooks, html_articles, and pdfs?

README.md
```diff
@@ -129,10 +129,27 @@ There are Datasets defined for various types of data sources - first check if an
...
+When wishing to update the whole dataset, run `python main.py fetch_all`. You can also fetch a specific subsection of a dataset by its name, for example `python main.py fetch aisafety.info`
 
+### Coda
 
-When wishing to update the whole dataset, run `python main.py fetch_all`. You can also fetch a specific subsection of a dataset by its name, for example `python main.py fetch aisafety.info`
+To update the stampy portion of the dataset, you will need a Coda token. go to coda.io, log in, and generate an API token in your account settings. Add restrictions: Doc or table, Read only, for the doc with url https://coda.io/d/_dfau7sl2hmG. Then, create a .env file at the root of the alignment research dataset, and write CODA_TOKEN="<coda_token>". It will be accessible in align_data/stampy/stampy.py
+
+### Google API
```
ccstan99, line 138: Ugggh, sorry. Didn't realize google sheets was going to make this so horrible for you.

ccstan99 (dismissed):
> python main.py update_metadata ... special_docs.csv ...

Should those long keys be in a constant in our file or are these credentials unique for each user?

The html_articles on HF look fine. I uploaded pdfs.jsonl & ebooks.jsonl to HF but can't seem to inspect them in dataset viewer.

.github/workflows/push-datasets.yml
mruwnik: whoops - good catch. This should be removed

mruwnik:
I dunno about the strings. It's a matter of whether the special_docs.csv should be viewed as the main source of truth, or whether in theory any old worksheet can be used.

The pdfs seem to work on HF now, the ebooks won't be shown as I disabled them for copyright reasons

align_data/gdocs/__init__.py
```diff
@@ -4,6 +4,5 @@
 # from https://www.lesswrong.com/posts/LfHWhcfK92qh2nwku/transcripts-of-interviews-with-ai-researchers
 
 GDOCS_REGISTRY = [
-    Gdocs(name = "gdocs",
-        gdrive_address = "https://drive.google.com/uc?id=18uFLj3Vs8de6LnEE00taJAvPl8dZYRxx")
+    Gdocs(name="gdocs", gdrive_address="https://drive.google.com/uc?id=18uFLj3Vs8de6LnEE00taJAvPl8dZYRxx"),
```
ccstan99, line 7: Do we need to move gdocs move to articles since metadata needs to be manually encoded? Or did you figure out how to deal with that already?

.github/workflows/fetch-dataset.yml
```diff
@@ -21,21 +21,23 @@ on:
...
-          - miri
           - jsteinhardt_blog
           - lesswrong
           - markdown.ebooks
+          - miri
           - ml_safety_newsletter
           - nonarxiv_papers
           - qualiacomputing
+          - pdfs
           - reports
```
ccstan99, line 39: Will we still have nonarxiv and reports or do they essentially become part of pdfs? I know we're probably still missing some that couldn't be processed.

ccstan99 (dismissed):
I'm thinking hardcode special_docs.csv as default but users can specify another file id to override? Seems like what you did with PDFS_FOLDER_ID read from .env?

Ah, makes sense with copyright. So for stampy-chat, should they download ebooks.jsonl from gdrive separately from the ARD on HF?

align_data/gdocs/__init__.py
mruwnik: I thought it would be cleaner to keep them separate:
* `articles` - does black magic to populate sheets with things to process
* `Gdocs` - your average dataset thingy, which provides everything needed to read (and only read!) from a google sheet

.github/workflows/fetch-dataset.yml
mruwnik: essentially yes. I reckon it'll be safer to do that later in a separate issue, as this is already large enough

mruwnik:
I'm not sure what the best way would be to provide the ebooks to the chat bot :/ 
Well, the best way would be to just get permission to add them. Other than that, the chatbot will have to manually fetch them, or maybe even have them on disk when deployed? That might be the best option, as adding them to the github repo would probably also be a bad idea. Seems worth discussing during the meeting

PR #64, opened 2023-06-30
Fetch all greaterwrong items

mruwnik:


align_data/greaterwrong/greaterwrong.py
```diff
@@ -145,14 +145,13 @@ def items_list(self):
...
         while next_date:
             posts = self.fetch_posts(self.make_query(next_date))
+            if not posts['results']:
+                return
 
             for post in posts['results']:
                 if post['htmlBody'] and self.tags_ok(post):
                     yield post
 
-            if len(posts['results']) < 50:
```
mruwnik, line 153: sometimes graphql returns fewer than the requested items. I'm guessing it does some post processing filtering stuff. This way there will be an additional fetch, but seeing are there are thousands of items, it won't really be that noticeable, and will at least fetch everything. Unless there is a very unlucky gap...

main.py
```diff
@@ -30,25 +30,27 @@ def list(self) -> List[str]:
         """Returns a list of all the datasets"""
         return sorted(ALL_DATASETS)
 
-    def fetch(self, name, rebuild=False) -> None:
+    def fetch(self, *names, rebuild=False) -> None:
```
Thomas-Lemoine, line 33: This is smart and helpful :p

Thomas-Lemoine (approved):

PR #65, opened 2023-07-01
default set to empty string

Thomas-Lemoine:
replaced 'n/a' with '' essentially. + minor nits


align_data/arbital/arbital.py
```diff
@@ -123,7 +123,7 @@ def process_entry(self, alias):
             summary, text = extract_text(page['text'])
 
             return DataEntry({
-                'title': page['title'] if 'title' in page else 'n/a',
+                'title': page.get('title'),
```
mruwnik, line 126: this should probably be `page.get('title') or ''`, as otherwise you might get `None`

mruwnik (approved):

align_data/arbital/arbital.py
Thomas-Lemoine: Oh thx, I thought the default was '', I can also do page.get('title', '') to set the default

mruwnik: yup. Though that has an interesting edge case:
```
{'ble': 123}.get('bla', '') => ''
{'bla': 123}.get('bla', '') => 123
{'bla': None}.get('bla', '') => None
```

mruwnik (approved):

PR #67, opened 2023-07-02
Ids based on id_fields

Thomas-Lemoine:
When creating a DataEntry object, you can now add an optional field, id_fields, which has a default value of ['source', 'text'] currently, and which will use the elements of the id_fields list (concatenated) to create the string that will then be hashed to yield the DataEntry's id. This means that stampy, for example, has the dataentry section changed to:
```
return DataEntry({
    "source": self.name,
    "source_type": "markdown",
    "url": url,
    "title": question,
    "authors": ['Stampy aisafety.info'],
    "date_published": self._get_published_date(entry),
    "text": answer,
}, id_fields=['url'])
```
Which means stampy now uses the 'url' only for id_fields. All sources are then easy to modify this way. One glaring problem is that the 'url' in fact does not uniquely describe stampy entries, since some entries have the same url (something to investigate)


Thomas-Lemoine:
I'm very much not certain this is the best way to do it, having an extra optional field in the initialization of DataEntry, so open to suggestions / alternatives, but this seems to work surprisingly well

PR #69, opened 2023-07-02
fix_duplicates_in_cold_takes

Thomas-Lemoine:
There was a bug where the dataset would go to https://www.cold-takes.com/, and first take all the urls from the "featured posts" section, and then take all the urls from the "all posts" section, and so the featured posts were all duplicated twice in the cold_takes dataset. Now I ignore the featured posts and the duplicates are gone.

### DIFF
--- a/align_data/blogs/blogs.py
+++ b/align_data/blogs/blogs.py
@@ -16,7 +16,7 @@ logger = logging.getLogger(__name__)
 @dataclass
 class ColdTakes(HTMLDataset):
     title_selector = 'h2'
-    item_selector = ['article']
+    item_selector = 'div.post-feed article'
 
     cleaner = utils.HtmlCleaner(
         ["You might also like\.\.\..*", "\\n+", "\#\# Create your profile.*", "\n\xa0Comment/discuss\n", '\nClick lower right to download or find on Apple Podcasts, Spotify, Stitcher, etc.\n'],
@@ -52,7 +52,7 @@ class ColdTakes(HTMLDataset):
 
 class GenerativeInk(HTMLDataset):
     title_selector = 'h3'
-    item_selector = ['div', {'class': 'post'}]
+    item_selector = 'div.post.on-list'
 
     def _get_published_date(self, contents):
         possible_date_elements = [
--- a/align_data/blogs/blogs.py
+++ b/align_data/blogs/blogs.py
@@ -7,9 +7,6 @@ from align_data.common.html_dataset import HTMLDataset, RSSDataset
 from datetime import datetime, timezone
 from dateutil.parser import parse
 
-import logging
-logger = logging.getLogger(__name__)
-
 @dataclass
 class ColdTakes(HTMLDataset):
     title_selector = 'h2'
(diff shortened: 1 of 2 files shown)


align_data/blogs/blogs.py
```diff
@@ -7,6 +7,12 @@
...
 
+import requests
+from bs4 import BeautifulSoup
+
+import logging
+logger = logging.getLogger(__name__)
+
 @dataclass
 class ColdTakes(HTMLDataset):
     title_selector = 'h2'
```
mruwnik, line 18: wouldn't changing `item_selector` to  `['div.post-feed article']` do the same thing, without having to override `items_list`?

Thomas-Lemoine: I did not quite understand how item_selector and all that worked, but that may be the case, that's great I will test it

Thomas-Lemoine: Ok so I tried a few things. 'div.post-feed article' brings the # of entries found to 0, so it seems to be breaking it. I've gone back and forth with gpt4, and it seems like based on the html it thinks .findall(*self.item_selector) won't cut it, and there needs to be the following change in HTMLDataset, the parent class of ColdTakes:
```
@dataclass
class HTMLDataset(AlignmentDataset):
    ...
    container_selector: str = None  # Add this line
    ...
    @property
    def items_list(self):
        logger.info(f"Fetching entries from {self.url}")
        response = requests.get(self.url, allow_redirects=True)
        soup = BeautifulSoup(response.content, "html.parser")
        
        # Add these lines
        if self.container_selector:
            soup = soup.find(*self.container_selector)
        
        articles = soup.find_all(*self.item_selector)
        logger.info(f"Found {len(articles)} articles")
        return articles
```
Where it would initialize to None and, in ColdTakes, it would get the value of 
self.container_selector = ['div', {'class': 'post-feed'}]
which, come to think of it, looks a lot like the item_selector of GenerativeInk, ['div', {'class': 'post'}]...

Thomas-Lemoine: I looked into this for a while and the issue is that HTMLDataset assumes that we can only select a single class thing, like a div with class 'post', and find all of these in the html, and those will contain a 'a' thing that contains the url we're looking for.
but this is not the case for cold_takes: the 'article' class is in two different sections, a featured-posts section and a post-feed section. This means that find_all, alone, can't differentiate between the two sections, and so there are duplicated articles.

Here is a solution I found so far:
```
    import requests
    from bs4 import BeautifulSoup
    urls = ['https://www.cold-takes.com/', 'https://generative.ink/posts/']
    selectors = ['div.post-feed article', 'div.post.on-list']
    for url, selector in zip(urls, selectors):
        response = requests.get(url, allow_redirects=True)
        soup = BeautifulSoup(response.content, "html.parser")

        posts = soup.select(selector)


        for post in posts:
            link = post.select_one('a')['href']
            print(link)
        print(len(posts))
```
this successfully prints the urls for each of generative.ink and cold_takes. it uses select instead of find_all.

Thomas-Lemoine: I made this work, updated the PR

Thomas-Lemoine:
made it neater by removing irrelevant changes

align_data/blogs/blogs.py
mruwnik: yup, that makes things a lot clearer. I did it this way before I found out about the `select()` function

mruwnik (approved):

PR #71, opened 2023-07-02
Custom_id_from_id_fields

Thomas-Lemoine:
When creating a DataEntry object, you can now add an optional field, id_fields, which has a default value of ['source', 'text'] currently, and which will use the elements of the id_fields list (concatenated) to create the string that will then be hashed to yield the DataEntry's id. This means that stampy, for example, has the dataentry section changed to:

return DataEntry({
    "source": self.name,
    "source_type": "markdown",
    "url": url,
    "title": question,
    "authors": ['Stampy aisafety.info'],
    "date_published": self._get_published_date(entry),
    "text": answer,
}, id_fields=['url'])
Which means stampy now uses the 'url' only for id_fields. All sources are then easy to modify this way. One glaring problem is that the 'url' in fact does not uniquely describe stampy entries, since some entries have the same url (something to investigate)

### DIFF
--- a/align_data/analysis/analyse_jsonl_data.py
+++ b/align_data/analysis/analyse_jsonl_data.py
@@ -23,10 +23,12 @@ def validate_data(data_dict, seen_urls):
     # TODO: add more checks here
 
 def check_for_duplicates(data_dict, seen_urls):
-    if data_dict.get('url') in seen_urls:
-        raise ValueError(f"Duplicate URL found. \nUrl: {data_dict['url']}\nfirst_duplicate: {get_data_dict_str(seen_urls[data_dict['url']])}\nsecond_duplicate: {get_data_dict_str(data_dict)}\n\n\n\n")
+    id = data_dict.get('id')
+    if id in seen_urls:
+        #raise ValueError(f"Duplicate id found. \nId: {id}\nfirst_duplicate: {get_data_dict_str(seen_urls[id])}\nsecond_duplicate: {get_data_dict_str(data_dict)}\n\n\n\n")
+        seen_urls[id].append(data_dict)
     else:
-        seen_urls[data_dict['url']] = data_dict
+        seen_urls[id] = [data_dict]
 
     #TODO: Add more validation logic here
     return seen_urls 
@@ -35,7 +37,7 @@ def get_data_dict_str(data_dict):
     """
     Returns a string representation of the given data_dict.
     """
-    return f"source: {data_dict['source']}, title: {data_dict['title'][:50]}, date_pub: {data_dict['date_published']}, url: {data_dict['url']}"
+    return f"source: {data_dict['source']}, title: {data_dict['title'][:50]}, date_pub: {data_dict['date_published']}, url: {data_dict['url']}\n"
 
 def files_iterator(data_dir):
     """
@@ -55,6 +57,9 @@ def process_jsonl_files(data_dir):
             seen_urls = check_for_duplicates(data_dict, seen_urls)
         except ValueError as e:
             print(e)
+    for id, duplicates in seen_urls.items():
+        if len(duplicates) > 1:
+            print(f"{len(duplicates)} duplicate ids found. \nId: {id}\n{''.join([f'id {i+1}: {get_data_dict_str(duplicate)}' for i, duplicate in enumerate(duplicates)])}\n\n\n\n")
 
 def delete_all_txt_and_jsonl(data_dir):
     """
@@ -67,3 +72,4 @@ def delete_all_txt_and_jsonl(data_dir):
 
 if __name__ == "__main__":
     process_jsonl_files("data/")
+    #delete_all_txt_and_jsonl("data/")
--- a/align_data/analysis/analyse_jsonl_data.py
+++ b/align_data/analysis/analyse_jsonl_data.py
@@ -1,7 +1,8 @@
-import os
-import json
 from datetime import datetime
 from pathlib import Path
+import jsonlines
+
+from collections import defaultdict
 
 def is_valid_date_format(data_dict, format="%Y-%m-%dT%H:%M:%SZ"):
     """
@@ -24,8 +25,6 @@ def validate_data(data_dict):
 
 def check_for_duplicates(data_dict, seen_urls):
     id = data_dict.get('id')
-    if not id in seen_urls:
-        seen_urls[id] = []
     seen_urls[id].append(data_dict)
 
     #TODO: Add more validation logic here
@@ -43,12 +42,12 @@ def files_iterator(data_dir):
     and yields every element (which is a dictionary) in the jsonl file.
     """
     for path in Path(data_dir).glob('*.jsonl'):
-        with open(path, encoding='utf-8') as f:
+        with jsonlines.open(path) as f:
             for line in f:
-                yield json.loads(line)
+                yield line
 
 def process_jsonl_files(data_dir):
-    seen_urls = dict()  # holds all seen urls
+    seen_urls = defaultdict(list)  # holds all seen urls
     for data_dict in files_iterator(data_dir):
         try:
             validate_data(data_dict)
@@ -58,8 +57,7 @@ def process_jsonl_files(data_dir):
         except Exception as e:
             print(f"Unexpected error: {e}")
     dup_count = 0
-    print(len(seen_urls))
-    print(type(seen_urls))
+    
     for id, duplicates in seen_urls.items():
         if len(duplicates) > 1:
             list_of_duplicates = '\n'.join(get_data_dict_str(duplicate) for duplicate in duplicates)
--- a/align_data/analysis/analyse_jsonl_data.py
+++ b/align_data/analysis/analyse_jsonl_data.py
@@ -14,7 +14,7 @@ def is_valid_date_format(data_dict, format="%Y-%m-%dT%H:%M:%SZ"):
     except ValueError:
         return False
 
-def validate_data(data_dict, seen_urls):
+def validate_data(data_dict):
     """
     Processes each dictionary element in the jsonl file. 
     """
@@ -24,11 +24,9 @@ def validate_data(data_dict, seen_urls):
 
 def check_for_duplicates(data_dict, seen_urls):
     id = data_dict.get('id')
-    if id in seen_urls:
-        #raise ValueError(f"Duplicate id found. \nId: {id}\nfirst_duplicate: {get_data_dict_str(seen_urls[id])}\nsecond_duplicate: {get_data_dict_str(data_dict)}\n\n\n\n")
-        seen_urls[id].append(data_dict)
-    else:
-        seen_urls[id] = [data_dict]
+    if not id in seen_urls:
+        seen_urls[id] = []
+    seen_urls[id].append(data_dict)
 
     #TODO: Add more validation logic here
     return seen_urls 
@@ -53,13 +51,21 @@ def process_jsonl_files(data_dir):
     seen_urls = dict()  # holds all seen urls
     for data_dict in files_iterator(data_dir):
         try:
-            validate_data(data_dict, seen_urls)
+            validate_data(data_dict)
             seen_urls = check_for_duplicates(data_dict, seen_urls)
         except ValueError as e:
             print(e)
+        except Exception as e:
+            print(f"Unexpected error: {e}")
+    dup_count = 0
+    print(len(seen_urls))
+    print(type(seen_urls))
     for id, duplicates in seen_urls.items():
         if len(duplicates) > 1:
-            print(f"{len(duplicates)} duplicate ids found. \nId: {id}\n{''.join([f'id {i+1}: {get_data_dict_str(duplicate)}' for i, duplicate in enumerate(duplicates)])}\n\n\n\n")
+            list_of_duplicates = '\n'.join(get_data_dict_str(duplicate) for duplicate in duplicates)
+            print(f"{len(duplicates)} duplicate ids found. \nId: {id}\n{list_of_duplicates}\n\n\n\n")
+            dup_count += 1
+    print(f"Total number of duplicate ids found: {dup_count}")
 
 def delete_all_txt_and_jsonl(data_dir):
     """
--- a/align_data/analysis/analyse_jsonl_data.py
+++ b/align_data/analysis/analyse_jsonl_data.py
@@ -14,28 +14,54 @@ def is_valid_date_format(data_dict, format="%Y-%m-%dT%H:%M:%SZ"):
     except ValueError:
         return False
 
-def validate_data(data_dict):
+def validate_data(data_dict, seen_urls):
     """
     Processes each dictionary element in the jsonl file. 
     """
     if not is_valid_date_format(data_dict):
         raise ValueError(f"Invalid date format for source: {data_dict['source']}, title: {data_dict['title'][:30]}, date_pub: {data_dict['date_published']}")
     
-    #TODO: Add more validation logic here
+    if data_dict.get('url') in seen_urls:
+        raise ValueError(f"Duplicate URL found. \nUrl: {data_dict['url']}\nfirst_duplicate: {get_data_dict_str(seen_urls[data_dict['url']])}\nsecond_duplicate: {get_data_dict_str(data_dict)}\n\n\n\n")
+    else:
+        seen_urls[data_dict['url']] = data_dict
 
-def process_jsonl_files(data_dir):
+    #TODO: Add more validation logic here
+    return seen_urls 
+
+def get_data_dict_str(data_dict):
+    """
+    Returns a string representation of the given data_dict.
+    """
+    return f"source: {data_dict['source']}, title: {data_dict['title'][:60]}, date_pub: {data_dict['date_published']}, url: {data_dict['url']}"
+
+def files_iterator(data_dir):
     """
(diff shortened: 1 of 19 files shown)


align_data/analysis/analyse_jsonl_data.py
```diff
@@ -14,28 +14,54 @@ def is_valid_date_format(data_dict, format="%Y-%m-%dT%H:%M:%SZ"):
...
-                    validate_data(data_dict)
-                except ValueError as e:
-                    print(e)
+                yield json.loads(line)
+
+def process_jsonl_files(data_dir):
+    seen_urls = dict()  # holds all seen urls
+    for data_dict in files_iterator(data_dir):
+        try:
+            seen_urls = validate_data(data_dict, seen_urls)
```
mruwnik, line 52: it would probably be better to separate the validation and the updating of the seen urls dict - they're quite different operations

main.py
```diff
@@ -52,7 +52,7 @@ def fetch(self, *names, rebuild=False) -> None:
 
             print(dataset.jsonl_path)
 
-    def fetch_all(self, rebuild=False, skip='') -> str:
+    def fetch_all(self, *skip, rebuild=False) -> str:
```
mruwnik, line 55: nice

mruwnik (commented):
I'm not sure about this. How about discussing it during the meeting on tuesday first?

align_data/analysis/analyse_jsonl_data.py
Thomas-Lemoine: that's fair, I can change that. I also agree it's worth discussing tuesday

Thomas-Lemoine: done

align_data/analysis/analyse_jsonl_data.py
```diff
@@ -14,28 +14,56 @@ def is_valid_date_format(data_dict, format="%Y-%m-%dT%H:%M:%SZ"):
...
+            seen_urls = check_for_duplicates(data_dict, seen_urls)
+        except ValueError as e:
+            print(e)
+
+def delete_all_txt_and_jsonl(data_dir):
+    """
+    Deletes all txt and jsonl files in the given directory.
+    """
+    for path in Path(data_dir).glob('*.txt'):
+        os.remove(path)
```
mruwnik, line 64: you can use `path.unlink()`

align_data/analysis/analyse_jsonl_data.py
```diff
@@ -14,28 +14,56 @@ def is_valid_date_format(data_dict, format="%Y-%m-%dT%H:%M:%SZ"):
...
     """
     Processes each dictionary element in the jsonl file. 
     """
     if not is_valid_date_format(data_dict):
         raise ValueError(f"Invalid date format for source: {data_dict['source']}, title: {data_dict['title'][:30]}, date_pub: {data_dict['date_published']}")
-    
+    # TODO: add more checks here
+
+def check_for_duplicates(data_dict, seen_urls):
+    if data_dict.get('url') in seen_urls:
```
mruwnik, line 26: I was thinking more in the vein of the checking part being in `validate_data` and a small `get_urls(data_dict, seen_url)` function that updates the `seen_urls` dict with the ones from the file. Though this does the job fine

align_data/analysis/analyse_jsonl_data.py
Thomas-Lemoine: I change these functions a lot in the write_tests branch I haven't yet published... where it's kind of messy tbh. Not sure what to do here, maybe that write_tests branch should be based on this branch and then I can pr that branch onto this one, not sure

align_data/common/alignment_dataset.py
```diff
@@ -252,24 +252,32 @@ def folder_from_gdrive(self, url=None, output=None):
...
                 self[k] = default and default()
+        # Store id_fields in a way that does not interfere with UserDict's functionality
+        self.__id_fields = id_fields or ["source", "text"]
 
     def add_id(self):
-        assert self["text"] is not None, "Entry is missing text"
-        text_excerpt = self["text"][:TEXT_LEN].encode("utf-8")
-        self["id"] = hashlib.md5(text_excerpt).hexdigest()
+        for field in self.__id_fields:
+            assert self[field] is not None, f"Entry is missing {field}"
```
mruwnik, line 265: how about 
```
missing = [field in self.__id_fields if not self.get(field)]
assert not missing, f'Entry is missing the following fields: {missing}'
```
It's annoying to have to fix fields one by one if multiple ones are missing

align_data/common/alignment_dataset.py
```diff
@@ -252,24 +252,32 @@ def folder_from_gdrive(self, url=None, output=None):
...
+
+        id_string = ''.join(self[field] for field in self.__id_fields)[:TEXT_LEN].encode("utf-8")
+        assert id_string, "Entry has empty id_fields"
```
mruwnik, line 268: this could also fail for type related reasons - you could be extra fancy and make sure all the id fields are strings (or cast them to strings, which would probably be better)

align_data/common/alignment_dataset.py
```diff
@@ -252,24 +252,32 @@ def folder_from_gdrive(self, url=None, output=None):
...
         assert self["id"] is not None, "Entry is missing id"
-        assert self["text"] is not None, "Entry is missing text"
-        text_excerpt = self["text"][:TEXT_LEN].encode("utf-8")
-        assert self["id"] == hashlib.md5(
-            text_excerpt).hexdigest(), "Entry id does not match text"
+        for field in self.__id_fields:
+            assert self[field] is not None, f"Entry is missing {field}"
+
+        id_string = ''.join(self[field] for field in self.__id_fields)[:TEXT_LEN].encode("utf-8")
+        assert id_string, "Entry has empty id_fields"
```
mruwnik, line 278: how about moving these to a `verify_fields()` method or something, seeing as `add_id` also uses them?

align_data/common/html_dataset.py
```diff
@@ -83,7 +82,7 @@ def process_entry(self, article):
             "date_published": date_published,
             "authors": self.extract_authors(contents),
             **self._extra_values(contents),
-        })
+        }, id_fields=["url"])
```
mruwnik, line 85: it might be worth having this as a class property of AlignmentDataset and also updating the base `AlignmentDataset.process_entry()` method to also add it

mruwnik (commented):
sorry for the radio silence

align_data/common/alignment_dataset.py
Thomas-Lemoine: To be clear, this would be         
```
missing = [field for field in self.__id_fields if self.get(field) is not None]
assert not missing, f'Entry is missing the following fields: {missing}'
```
right? If the field is empty, like it has an empty url (which some sources do afair), that is fine as long as the cumulative string made up of their url and title and whatever else, to build the id, is not empty

mruwnik: right

align_data/analysis/analyse_jsonl_data.py
```diff
@@ -14,28 +14,62 @@ def is_valid_date_format(data_dict, format="%Y-%m-%dT%H:%M:%SZ"):
...
             for line in f:
-                data_dict = json.loads(line)
-                try:
-                    validate_data(data_dict)
-                except ValueError as e:
-                    print(e)
+                yield json.loads(line)
+
+def process_jsonl_files(data_dir):
+    seen_urls = dict()  # holds all seen urls
```
mruwnik, line 53:
```
from collections import defaultdict

seen_urls = defaultdict(list)

seen_urls['bla'].append(1)
```

align_data/analysis/analyse_jsonl_data.py
```diff
@@ -14,28 +14,62 @@ def is_valid_date_format(data_dict, format="%Y-%m-%dT%H:%M:%SZ"):
...
+    return f"source: {data_dict['source']}, title: {data_dict['title'][:50]}, date_pub: {data_dict['date_published']}, url: {data_dict['url']}\n"
+
+def files_iterator(data_dir):
     """
     Goes through the data directory, opens every jsonl file sequentially, 
-    and processes every element (which is a dictionary) in the jsonl file.
+    and yields every element (which is a dictionary) in the jsonl file.
     """
     for path in Path(data_dir).glob('*.jsonl'):
         with open(path, encoding='utf-8') as f:
```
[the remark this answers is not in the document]
mruwnik: Or you can use `AlignmentDataset.read_entries()` which is pretty much the same as this function

align_data/common/alignment_dataset.py
Thomas-Lemoine: done

align_data/common/html_dataset.py
Thomas-Lemoine: The tricky part is that the whole adding ids and using id_fields currently happens in DataEntry, and AlignmentDataset's process_entry method raises NotImplementedError, and so I'm not sure how to give a field to AlignmentDataset so that, without modifying the code in all the process_entries methods, we can, for all different sources, create a DataEntry not dependant on the class of the source itself but also have it gain the id_fields from the field of the dataset in question? I'm not making a lot of sense, but what I mean is that if
```
@dataclass
someclass(alignmentdataset)
 self.__id_fields
 def process_entries(self,bla):
     return DataEntry({blabla})
```
is the current definition of a DataEntry in some specific source's dataset's process_entries method, I don't see how DataEntry can use the id_fields field of someclass, without modifying what DataEntry receives. And if we're gonna change what DataEntry receives anyway, it seems like it's clearer to write it out there only?

Thomas-Lemoine: Perhaps someclass would have a __id_fields field, and so we can fill it (but it has the 'url' default), and then all DataEntry objects we create have, at the end of them, added the
```
return DataEntry({blabla}, id_fields = __id_fields)
```

so that we can define the id_fields from the __init__.py files so they're more easily accessible

align_data/common/alignment_dataset.py
```diff
@@ -258,24 +258,35 @@ def folder_from_gdrive(self, url=None, output=None):
...
+
+    def generate_id_string(self):
+        return ''.join(str(self[field]) for field in self.__id_fields).encode("utf-8")
 
+    def verify_fields(self):
+        missing = [field for field in self.__id_fields if self.get(field) is None]
+        assert not missing, f'Entry is missing the following fields: {missing}'
+
+        id_string = self.generate_id_string()
+        assert id_string, "Entry has empty id_fields"
```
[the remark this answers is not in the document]
Thomas-Lemoine: No, it means that ''.join(self[id_field] for id_field in __id_fields) is == '', ie the string that is getting hashed for the id is an empty string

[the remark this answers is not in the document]
Thomas-Lemoine: I can make it clearer by saying assert id_string != '', for example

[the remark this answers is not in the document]
Thomas-Lemoine: and making the error message less ambiguous, I agree that right now it is needlessly confusing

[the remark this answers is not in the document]
mruwnik: why not just change `missing = [field for field in self.__id_fields if self.get(field) is None]` to `missing = [field for field in self.__id_fields if not self.get(field)]`? That should then handle both cases. Or do you want to differentiate between a key being absent and it being empty?

align_data/analysis/analyse_jsonl_data.py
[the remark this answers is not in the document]
Thomas-Lemoine: I agree jsonlines should work but `import jsonlines` has been saying
`Import "jsonlines" could not be resolvedPylance[reportMissingImports](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportMissingImports)` for a while and I don't exactly get why, since I pip installed it, and it is in the pip list

align_data/common/alignment_dataset.py
```diff
@@ -258,24 +258,35 @@ def folder_from_gdrive(self, url=None, output=None):
...
+        return id_string
+        
     def add_id(self):
-        assert self["text"] is not None, "Entry is missing text"
-        text_excerpt = self["text"][:TEXT_LEN].encode("utf-8")
-        self["id"] = hashlib.md5(text_excerpt).hexdigest()
+        id_string = self.verify_fields()
```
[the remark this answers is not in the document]
Thomas-Lemoine: I agree a bit, but the alternative I had was to compute the id_string twice for no reason, unless I missed something.

align_data/analysis/analyse_jsonl_data.py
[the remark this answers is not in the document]
mruwnik: is that an editor error, or something that happens when you run it?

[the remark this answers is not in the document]
Thomas-Lemoine: the issue's fixed, it was a multiple-interpreters thing apparently

align_data/common/alignment_dataset.py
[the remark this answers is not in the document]
Thomas-Lemoine: The reason I did it that way is to be flexible to entries that would, say, not have a url, but have a unique title; as long as one of the id_fields was a non empty string the id will be unique. If we do it that way, any empty field will cound the entry as having missing fields, which is also fine but might trigger more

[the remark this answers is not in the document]
mruwnik: shouldn't all the id fields have a value, though?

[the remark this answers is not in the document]
Thomas-Lemoine: Yes, I suppose I designed it that way back when I thought it may be reasonable for id fields to include like 5 fields in which case it seemed relevant to ensure flexibility or something, but you're right that they should so that will simplify the code

align_data/analysis/analyse_jsonl_data.py
```diff
@@ -20,22 +21,59 @@ def validate_data(data_dict):
...
-                data_dict = json.loads(line)
-                try:
-                    validate_data(data_dict)
-                except ValueError as e:
-                    print(e)
+                yield line
+
+def process_jsonl_files(data_dir):
+    seen_urls = defaultdict(list)  # holds all seen urls
+    for data_dict in files_iterator(data_dir):
```
mruwnik, line 51: what if it tries to open a file that has invalid json lines in it? Won't it just blow up here? (that can be desired - I just want to know that you're aware of this)

align_data/common/html_dataset.py
mruwnik: yup, this exactly

align_data/analysis/analyse_jsonl_data.py
Thomas-Lemoine: I did not consciously set it out to be that way but it's always good when code breaks when it doesn't get what you expect it to get, or maybe not always but sometimes, and so yeah I guess I'd be happy to see it blow up in the case of a malfunction of the dataset creation process

Thomas-Lemoine:
I see this newest commit doesn't quite work bc it won't let me do 
```
STAMPY_REGISTRY = [
    Stampy(name='aisafety.info', id_fields=['url']),
]
```, not sure why yet but it should be an easy fix

File: align_data/analysis/analyse_jsonl_data.py (hunk 13 above)
Comment (mruwnik) [reply align_data/analysis/analyse_jsonl_data.py:51]: cool. In that case I'll approve this once you resolve the conflicts :D

**Comment by Thomas-Lemoine** [comment]
> I see this newest commit doesn't quite work bc it won't let me do
> 
> ```
> STAMPY_REGISTRY = [
>     Stampy(name='aisafety.info', id_fields=['url']),
> ]
> ```, not sure why yet but it should be an easy fix
> ```

probably a dataclass thing

**Comment by Thomas-Lemoine** [comment]
ok I went in circles until I found out about KW_ONLY, and now I'm still kind of going in circles but hopefully less.
@mruwnik 
so right now the plan is to have 
```py
class Alignmentdataset:
    # blabla
    id_fields: List[str] = field(default_factory=lambda: ['url', 'title'])
    
    # blabla
    def make_data_entry(self, data, **kwargs):
        return DataEntry(dict(data, **kwargs), id_fields=self.id_fields)
    # blabla
    
class DataEntry(UserDict):
    def __init__(self, *args, id_fields=None, **kwargs):
        super().__init__(*args, **kwargs)
        for k, default in INIT_DICT.items():
            if k not in self:
                self[k] = default and default()
        # Store id_fields in a way that does not interfere with UserDict's functionality
        self.__id_fields = id_fields
```
and the problem seems to be that in the tests we use DataEntry directly. is it weird to have the "default value" of ['url', 'title'] be defined twice in two different places? I could set it to a default in AlignmentDataset, and then a default in DataEntry which realistically would only be used if it was created independently of AlignmentDataset, which it should only do during tests. It seems a bit suboptimal, so do you reckon it's fine or is there a better way you can think of

Thomas-Lemoine:
Hmmm I see, so now DataEntry is kind of a private class of AlignmentDataset, and I change the tests to reflect that

mruwnik (approved):
You'll have to:
```
git checkout main
git pull
git checkout -
git rebase main
```
fix the conflicts, then
```
git rebase --continue
git push -f
```

Thomas-Lemoine:
ok done, but the conflicts means you gotta approve it again

mruwnik (approved):
Make sure you choose 'squash and merge'

PR #72, opened 2023-07-03
WordPressBlog's tqdm process bar now shows correct total number of pages

Thomas-Lemoine:
NITS
Previously, when processing a wordpress blog like yudkowsky.net, tqdm would show the process bar as
1/2000 
2/2000
...
33/2000
and then be done, since there were actually just 33 pages in yudkowsky.net. the 2000 value was from assuming a very high page number (max_pages = 2000) and breaking the loop when we find a page that is empty. 

This PR uses exponential search to check that the page 1, 2, 4, 8, 16, etc exists, and then uses binary search on the bounds (for example, we determine the page number is between 64 and 128, so those are the bounds for our binary search), and so gets, in ~2-10 requests, an accurate number. It also will not break if a wordpress blog happens to have 2001 pages or more, and make the user aware of this fact quickly.

Now, all sources have an accurate display bar. An alternative is to go the greaterwrong.py route and not show a display bar but instead have self.items_list be an iterable, but that gives the user no idea how long it will take to process every entry (it might still be the better solution). Additionally, this PR feature never slows the fetching down by more than a few seconds.

🤷‍♂️ might be useless idk, it just nagged me a bit


mruwnik:
it's not useless. I also considered doing something like this, as it also bothered me :D
The problem with it is that it slows things down. On the other hand, it's only an extra `log(n)` calls...
One thing you could consider would be to cache the results of the test calls, especially as most of the sites aren't going to have more than a couple of pages

mruwnik:
Another option would be to set `lazy_eval = True` and change `items_list` to return a generator, i.e.:
```
@dataclass
class WordpressBlog(AlignmentDataset):
    url: str
    (...)
    lazy_eval = True

    (...)

    @property
    def items_list(self):
        return (f"{self.feed_url}?paged={page + 1}" for page in range(0, self.max_pages))
```
This will just show the number of items processed per second, rather than a proper progress bar

align_data/blogs/wp_blog.py
```diff
@@ -37,7 +37,46 @@ def get_item_key(self, item):
...
+            mid = (lower_bound + upper_bound) // 2
+            if is_valid_page(mid):
+                lower_bound = mid
+            else:
+                upper_bound = mid
+            logger.info('lower_bound', lower_bound, 'upper_bound', upper_bound)
+
+        # add the valid pages to pages_url
+        for i in range(lower_bound + 1):  # Adding 1 to include the last valid page
+            pages_url.append(f"{self.feed_url}?paged={i + 1}")
```
mruwnik, line 77: If you moved the checking to a function (e.g. `get_page_count(self)`, this could be done as
```
    @property
    def items_list(self):
        return [f"{self.feed_url}?paged={page + 1}" for page in range(self.get_page_count())]
```

align_data/blogs/wp_blog.py
```diff
@@ -37,7 +37,46 @@ def get_item_key(self, item):
...
+        which makes tqdm think that there are 2000 items 
+        to process, we'll find the last valid page number 
+        and create a list of urls up to that page number.
+        We use exponential search to find the last valid page number.
+        """
+        last_title = [None]  # We'll use a list so that we can modify it inside the inner function
+
+        def is_valid_page(i):
+            paged_url = f"{self.feed_url}?paged={i + 1}"
+            d = feedparser.parse(paged_url)
```
mruwnik, line 51: you could probably get away with `return requests.head(paged_url).status_code == 200`

Thomas-Lemoine: looks like status_code always returns 403, no matter if the page should exist or not, unless my testing is wrong

Thomas-Lemoine: Apparently that means I'm forbidden from HEAD requests or wtv. bc of that gpt4 opted for using beautiful soup instead of requests.head but then isn't that defeating the purpose of switching away from the parser, since afaik it's getting the whole page either way which presumably is the reason requests.head() should be quicker

align_data/blogs/wp_blog.py
Thomas-Lemoine: done

Thomas-Lemoine: and yeah to the previous comment (to which I can't reply ig) that also works, it's really a tradeoff between a slightly neater tqdm load bar and ~5-10 seconds slower, or greaterwrong-like lazy_eval. I could add an option to switch between the two but not sure it's worth it. If you have a preference for using lazy_eval with a generator then that's fine by me

align_data/blogs/wp_blog.py
mruwnik: wordpress checks the user agents header and doesn't let you in if it thinks you're a bot. Try (after pulling the newset main):
```
from align_data.articles.parsers import fetch
return fetch(paged_url, method='head').status_code == 200
```

mruwnik:
One other possible approach, if you don't mind a longer initial wait, and don't expect to see massive wordpress sites (as in like GBs worth of text) would be to just fetch all the RSS sub pages into memory. Then you could do something like:
```

@dataclass
class WordpressBlog(RSSDataset):
    url: str
    summary_key = 'summary'
    done_key = 'paged_url'
    max_pages = 2000

    @property
    def items_list(self):
        logger.info(f"Fetching entries from {self.feed_url}")

        pages = []
        for page in range(self.max_pages):
            url = f"{self.feed_url}?paged={page + 1}"
            feed = feedparser.parse(self.feed_url)
            self.items.update({item['link']: item for item in feed['entries']})
            pages += list(self.items.keys())

        logger.info(f'Got {len(pages)} pages')

        return pages
```
and just accept the initial slowdown

Thomas-Lemoine:
right that makes sense, instead of getting an item_list where the items are page links, we just get the pages themselves into a list

align_data/blogs/wp_blog.py
Thomas-Lemoine: align_data.articles? is that a thing?

Thomas-Lemoine:
One issue with that is that then around 90% of the time will be spent before the process bar is even shown which kind of defeats the purpose of a process bar

Thomas-Lemoine:
How about this change?

align_data/blogs/wp_blog.py
mruwnik: it is now...

align_data/blogs/wp_blog.py
```diff
@@ -16,28 +15,44 @@
 class WordpressBlog(AlignmentDataset):
```
mruwnik, line 15: if you go this route, I'd prefer it as a subclass of RSSDataset, as that's what it really is

mruwnik:
This comes down to whether it's an issue to wait a long time. Unless you do 2 tmdq loops... You could do the one where it just shows items per second for the fetching of the data, then the basic one with percentages for the downloaded items :D 
Are we having fun yet? I really hope you're not getting annoyed at all my bugging

align_data/blogs/wp_blog.py
Thomas-Lemoine: Hmmm, am I supposed to frequently fetch the origin/main and then merge it into my branches btw? Or should it just be a only-when-useful thing

align_data/blogs/wp_blog.py
Thomas-Lemoine: Ah I see.

align_data/blogs/wp_blog.py
mruwnik: I'm guessing whatever you find useful. I try to always branch off the most current version of main and update my branches every few days. It depends a lot on how fast things are moving. Unless you're working on something that gets touched a lot by other people you should be fine only updating when needed. In this case it was something I merged in today, which was causing me problems, so the info was fresh in my mind.

Thomas-Lemoine:
Yeah two progress bars would make all the other sources look feeble in comparison 😎 so one when creating the items_list which is slowish, and one when processing all the entries which is fast. Wait but doesn't this run into the same issue, we still don't know how many pages we'll reach! it could be the progress bar from generators but that one isn't as slick.
One thing is that I have a final tomorrow for a class I'm neglecting so I should probably do this stuff only after the final.

Thomas-Lemoine:
I just realized I failed at reading comprehension and said the same thing I said but yes exactly two bars with the first one just showing the speed so at least the user knows something's going on and there's not just *silence*

Thomas-Lemoine:
and you said the same thing I said*

mruwnik:
could I also bother you to add tests for the WP blogs? If you'd prefer not to, I can do it later

Thomas-Lemoine:
Yeah I can. I'm wasting a lot of time on this branch rn, not really managing to make anything work that was better than the original commit Imo. RSSDataset is confusing to me currently.

mruwnik (dismissed):
Looks good. I'd still rather it as a RSSDataset, but I can do it later, or I can explain the code to you if you prefer

mruwnik (approved):

Thomas-Lemoine:
Yeah RSS would indeed be better I guess this is all silly in that case

mruwnik (approved):

Thomas-Lemoine:
LOL well I guess after troubleshooting a bunch of things with it I now see how similar it is to RSS

mruwnik:
Yup :D But you're right in it not being obvious. And your changes are really valuable in that they moved all the ugly bits into one place (the `items_list` property), which now allows this to be easily transformed into a RSSDataset

PR #77, opened 2023-07-06
tests for html datasets

mruwnik:
Tests!
This also fixes some small bugs I found along the way

### DIFF
--- a/align_data/articles/datasets.py
+++ b/align_data/articles/datasets.py
@@ -8,6 +8,7 @@ import requests
 import pypandoc
 import pandas as pd
 from gdown.download import download
+from markdownify import markdownify
 
 from align_data.articles.pdf import fetch_pdf, read_pdf, fetch
 from align_data.articles.parsers import HTML_PARSERS
@@ -22,6 +23,7 @@ class SpreadsheetDataset(AlignmentDataset):
     spreadsheet_id: str
     sheet_id: str
     done_key = "title"
+    source_filetype = None
 
     @property
     def items_list(self):
@@ -41,7 +43,7 @@ class SpreadsheetDataset(AlignmentDataset):
 
     @staticmethod
     def extract_authors(item):
-        return [author.strip() for author in item.authors.split(',')]
+        return [author.strip() for author in item.authors.split(',') if author.strip()]
 
     def process_entry(self, item):
         text = self._get_text(item)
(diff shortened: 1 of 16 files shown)


align_data/articles/datasets.py
```diff
@@ -85,6 +90,7 @@ def _get_text(item):
 
 class EbookArticles(SpreadsheetDataset):
 
+    source_filetype = 'epub'
```
ccstan99, line 93: Ah great, thanks for catching these! I noticed it which browsing the dataset but hadn't gotten around to mentioning it.

align_data/blogs/gwern_blog.py
```diff
@@ -41,14 +41,15 @@ def process_entry(self, post_href):
             return None
 
         # Some pages are returned as markdown, some as HTML, so handle both
-        if 'text/html' in article.headers.get('Content-Type'):
+        if 'text/html' in article.headers.get('Content-Type', ''):
```
ccstan99, line 44: I'm beginning to wonder if it's important having gwern_blog as it's own dataset for 7 entries as opposed to just treating it like a special_docs but I guess there was a new article last month. Anyway, it's already done so just food for those someday.

ccstan99 (approved):
Looks good on a quick skim. If someone else has time & inclination to review more carefully, please do so. Thanks!

Thomas-Lemoine (commented):
Very organized tests!

align_data/common/alignment_dataset.py
```diff
@@ -211,7 +209,8 @@ def process_entry(self, entry):
 
     @staticmethod
     def _format_datetime(date):
-        dt = date.astimezone(timezone.utc)
+        # Totally ignore any timezone info, forcing everything to UTC
+        dt = date.replace(tzinfo=pytz.UTC)
```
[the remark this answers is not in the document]
mruwnik: in the case of timestamps with proper timezones, this will be incorrect, as it will probably change the time of publication. From what I can see, most of them don't have timezone info, or are UTC. On the other hand, if you're in UTC+8 and do something like:
```
dt = parse('2023/01/01 12:34')
dt = dt.astimezone(timezone.utc)
dt.isoformat()
```
you'll get '2023-01-01 04:34', as `parse()` will return a datetime in your timezone, as the provided string doesn't specify it.

A better option would be to handle both naive and timezoned datetimes, but that's a nightmare I want to avoid. I'm guessing the time isn't really that important, and even if the date is off by a day or two that shouldn't be too bad, and it's more important for the timestamps to be consistent.

[This](https://xkcd.com/1883/) is the classic reference at this point.

PR #80, opened 2023-07-07
Update HF README on upload

mruwnik:
This will:
* add an option to fetch previous datasets when parsing - this is nice for e.g. lesswrong, as it will only fetch new items, rather than everything
* update the HF README when pushing files - otherwise this step is needed to be done manually each time, which sort of defeats the point

### DIFF
--- a/.github/workflows/fetch-dataset.yml
+++ b/.github/workflows/fetch-dataset.yml
@@ -57,7 +57,7 @@ jobs:
       run: pip install -r requirements.txt
 
     - name: Fetch dataset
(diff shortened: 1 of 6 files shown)


ccstan99 (approved):
Can you please also do the same for the private dataset
https://huggingface.co/datasets/StampyAI/ard-private

We might want to make sure `ebooks.jsonl` doesn't get pushed to the public dataset. Even though it's not part of downloaded as part of load_dataset, the file was still visible & downloadable from the "Files and Versions" tab, so I deleted it.

PR #88, opened 2023-07-12
Check special indices for new articles

mruwnik:
this handles the yellow items from [the spreadsheet](https://docs.google.com/spreadsheets/d/1pgG3HzercOhf4gniaqp3tBc3uvZnHpPhXErwHcthmbI/edit?pli=1#gid=1275815248). Blue rows should be handled separately, and the orange ones were too annoying

### DIFF
--- a/align_data/articles/articles.py
+++ b/align_data/articles/articles.py
@@ -5,14 +5,16 @@ from tqdm import tqdm
 
 from align_data.articles.google_cloud import iterate_rows, get_spreadsheet, get_sheet, upload_file, OK, with_retry
 from align_data.articles.parsers import item_metadata, fetch
+from align_data.articles.indices import fetch_all
 from align_data.settings import PDFS_FOLDER_ID
 
 
 logger = logging.getLogger(__name__)
 
 
-REQUIRED_FIELDS = ['url', 'title', 'source_type', 'date_published']
-OPTIONAL_FIELDS = ['source_url', 'authors', 'summary']
+# Careful changing these - the sheets assume this ordering
+REQUIRED_FIELDS = ['url', 'source_url', 'title', 'source_type', 'date_published']
+OPTIONAL_FIELDS = ['authors', 'summary']
 
 
 def save_pdf(filename, link):
@@ -45,7 +47,7 @@ def process_row(row, sheets):
         logger.error('missing keys: ' + ', '.join(missing))
         return
 
-    source_url = row.get('source_url') or row['url']
+    source_url = row.get('source_url')
     contents = item_metadata(source_url)
 
     if not contents or 'error' in contents:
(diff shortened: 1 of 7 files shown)


align_data/articles/articles.py
```diff
@@ -101,3 +104,29 @@ def update_new_items(source_spreadsheet, source_sheet, output_spreadsheet):
     source_sheet = get_sheet(source_spreadsheet, source_sheet)
     sheets = {sheet.title: sheet for sheet in get_spreadsheet(output_spreadsheet).worksheets()}
     return process_spreadsheets(source_sheet, sheets)
+
+
+def check_new_articles(source_spreadsheet, source_sheet):
+    """Goes through the special indices looking for unseen articles."""
+    source_sheet = get_sheet(source_spreadsheet, source_sheet)
+    current = {row.get('title'): row for row in iterate_rows(source_sheet)}
+    seen_urls = {url for item in current.values() for url in [item.get('url'), item.get('source_url')] if url}
```
ccstan99, line 113: The main issue I'm seeing is that special_docs gets populated with a bunch of entries that are probably already in ARD (arxiv papers, youtube videos, some lesswrong posts, even agentmodels.org) but we weren't keeping in special_docs. If we move them from this tab, it'll get repopulated again, so we should discuss the best workflow.

align_data/articles/parsers.py
```diff
@@ -196,6 +207,7 @@ def getter(url):
             '.LicenseInfo', '.ArticleIdentifierLinks', '.Banner', '.screen-reader-main-title', '.Publication'
         ]
     ),
+    'transformer-circuits.pub': error('not handled yet - same codebase as distill'),
```
ccstan99, line 210: Not sure about same codebase but same authors probably similar format for processing.

ccstan99 (approved):

align_data/articles/articles.py
ccstan99: Probably less of an issue when we have the DB set up... that'll solve all our problems (thinking wishfully)? We'll still need to decide about what goes into the special_docs and whatnot though.

mruwnik: yup, that's what I was thinking. Instead of checking for urls this way, a unique constraint on the id column (or columns) would handle most of this madness

align_data/articles/parsers.py
mruwnik: yup, it looks like it's the same. I couldn't be bothered to do it now, but wanted to note it down for whenever it gets done

align_data/articles/articles.py
mruwnik: one way around this is to just mark any duplicates as so i.e. set the status column to reflect that. Or just set it to 'ok' for duplicates, as they will then be ignored by any subsequent metadata checks. But yes, it's a problem :/ The DB should help a lot with it, though

PR #89, opened 2023-07-12
Move WP blogs to RSS

mruwnik:
10min of work, thanks to the previous changes


align_data/blogs/wp_blog.py
```diff
@@ -1,95 +1,52 @@
...
+                title = feed.get('feed', {}).get('title')
+                if not title or title == prev_title:
                     break
-                last_title = feed["feed"]["title"]
 
-                pages.extend({**entry, 'paged_url': paged_url} for entry in feed['entries'])
+                prev_title = feed["feed"]["title"]
                 page_number += 1
 
+                for item in feed['entries']:
```
Thomas-Lemoine, line 43: since you removed paged_url from the item's entries, it raises an error in align_data/common/alignment_dataset.py at line 100, since entry[self.done_key] looks for the entry['paged_url'] which doesn't exist. Maybe I'm running it wrong somehow but if not that is an error. 

my second concern is much more uncertain, but isn't get_item_key supposed to have each item have a unique key? This is why I had changed the get_item_key to link, since I knew the link would be unique to a specific item and knew the page_url would not. But maybe it's just saying that, if that page_url already is in the dataset, this entry definitely already is in the dataset and so it can be ignored?

mruwnik: maybe, maybe not. In most cases yes, but there a couple where that assumption doesn't hold:
* one of the pages couldn't be fetched (e.g. a network error), which would mean that it would never be retried
* the newest pages would be ignored, at least until enough were added for `page_url` to go forward by one

Thomas-Lemoine: I guess this shows my general confusion about the done_key: why not just use id everytime?

align_data/common/utils.py
```diff
@@ -1,79 +0,0 @@
-import bs4
```
Thomas-Lemoine, line 1: nice

align_data/blogs/wp_blog.py
Thomas-Lemoine: ^ prev comment took a while to send, had not seen your reply

mruwnik: it should use id everytime, really. The original AlignementDataset only had the `done_key` to check for uniqueness, so it's sort of stayed around. If you can simplify things, please do - I keep forgetting about it

Thomas-Lemoine: Sounds good. and in the slow tests I'll eventually get around to completing, there would presumably be some test that checks that the whole done_key and duplicates business is actually working as expected? since tbh I almost exclusively run code with --rebuild and in any case it's kind of hard to tell if it is working correctly otherwise

ccstan99: Not directly related to the PR but this conversation reminded we that we need to discuss about re-fetching urls when content has been updated. I'm thinking blog posts, stampy answers, but also arxiv papers may go through several revisions after they're published.

mruwnik:
you can push changes to other peoples branches, that's not a problem. A word of warning though - I tend to force push, which will overwrite any changes, so make sure you warn people when working on a common branch. In that case it's pretty much the same as working on main - pull before pushing and you should be fine

mruwnik:
ah. I only just noticed you pushed, after I force pushed with rebased changes. Which just goes to illustrate my point :D

Thomas-Lemoine (approved):

Issue #60, opened 2023-06-28
Write tests


Thomas-Lemoine:
Write tests to ensure that every source consistently outputs data in the required formats, using pytest. See tests/align_data/common/test_alignment_dataset.py for an example of this.

Some things to check: 
- All plausible date_published formats for a source are taken into account, and other ones default to the empty string. The date appears like 
- 'title', 'text', 'id', 'source' must not be empty strings or None
- Summary and authors must be lists, defaulted to an empty list that contains strings as elements
- etc.

Thomas-Lemoine:
@mruwnik Do you think the "analyse_jsonl_data.py" should be replaced by a test in the tests/ folder? Like, is the tests/ folder supposed to be pretty simple tests? bc I could have the "test" be to download the whole dataset, or perhaps only the first few entries from each source idk, and then do various tests to ensure every key from every entry from every source has the desired formats, but that sounds pretty lame because it would take about an hour to run, etc. so I don't want it to be run by default with the pytest.
If I understand correctly, pytest will only run the file automatically if it starts with test_, correct ? So I could have that be an optional test that is more thorough but slow, with a name that doesn't start with test_.

mruwnik:
pretty much, yes. There should be tests that check whether the code does what you expect it to do (unit tests), where you feed the functions with known, prepared content with known outputs - these are mainly to help when changing the code, so you know if you accidentally broke something else. These should be fast and can be run often.

Then you have integration or even end-to-end tests that run real queries etc, but which are only run sometimes. Though come to think of it, this might be better as part of the actual code. i.e. after scraping and mangling the data, there could be an extra (maybe optional?) step that ensures that the resulting jsonl files are correct. 

Pytest has various options to [mark specific tests](https://docs.pytest.org/en/6.2.x/example/markers.html#mark-examples) (in this case as slow). This can get complicated, but is very useful. You could then have a marker like [here](https://docs.pytest.org/en/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option) to only run slow tests when explicitly selected

Thomas-Lemoine:
Oh that looks pretty neat. Currently, there's a --rebuild flag for fetch and fetch all, and there could be a --slowtests that, after running the whole thing, runs a slow test going through what values each entry in each source has. But also, in the test_ files, there could be one that has functions marked for a slowtest. so if you add a --runslow flag to pytest, it runs those as well.

mruwnik:
[...]

PR #92, opened 2023-07-13
experimenting with slow tests

Thomas-Lemoine:
WIP, #60 I don't really know what I'm doing and am reconsidering whether slow tests is the way to go for this. 

The explicit goal of this branch is to create advanced / slow tests that can be run rarely, but will ensure that whatever output is being sent to HuggingFace has correct formats, no quirks, no duplicate ids, etc. If our fetched datasets all pass this and the pytests, we should be confident that the code works as expected.

Unfortunately all my analyse_jsonl_data.py code disappeared at some point and I don't remember where it was but I deleted a bunch of branches so the progress I had made on it appears to be gone.
Moreover, the whole pytest format seems suboptimal. prints don't show up on the screen while it is testing, so fetching the entire dataset (could take hours) without any feedback that it is not stuck in a loop is not practical.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,21 @@
+import pytest
+
+
+def pytest_addoption(parser):
+    parser.addoption(
+        "--runslow", action="store_true", default=False, help="run slow tests"
+    )
+
+
+def pytest_configure(config):
+    config.addinivalue_line("markers", "slow: mark test as slow to run")
+
+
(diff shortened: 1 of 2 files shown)


mruwnik:
you can run pytest with the `-s` option for it to display stuff right away

Thomas-Lemoine:
Ohhh thanks!

PR #93, opened 2023-07-13
change the id_fields for the sources that create id duplicates

Thomas-Lemoine:


Thomas-Lemoine:
turns out I was wrong and I (or someone) had already changed the stampy one

mruwnik (approved):

PR #97, opened 2023-07-19
implementing the UNIMPLEMENTED_PARSERS

Thomas-Lemoine:


Thomas-Lemoine:
I'm pushing this branch before adding meaningful things to it, to note down confusions

Thomas-Lemoine:
I don't quite understand why row.set_status set the status to '' at some point in the code because I don't see how such an issue can occur

Thomas-Lemoine:
I just figured out why fetching 'pdfs' and 'html' and etc. using the AlignmentDataset class seemed not to depend on the alignment-research-dataset-sources spreadsheet, https://docs.google.com/spreadsheets/d/1pgG3HzercOhf4gniaqp3tBc3uvZnHpPhXErwHcthmbI/edit?pli=1#gid=980957638. It's actually just getting the items from https://docs.google.com/spreadsheets/d/1l3azVJVukGAvZPgg0GyeqiaQe8bEMZvycBJaA8cRXf4/edit#gid=759210636, items-metadata, afaict.

Thomas-Lemoine:
@mruwnik gpt4 recommended I create a "logger_config.py" file that configures a logger for every file so that any file can just do from logger_config import logger, and then use the logger as before, with logger.info and whatever else, and it'll all show up in a log file that is gitignored but will print relevant info etc. Thoughts? maybe that clutters stuff a bit more or is more error prone or less readable etc.

mruwnik:
meh? Are you actually going to use the file? Currently this gets run on github actions, so all you get is what is printed to stdout (unless you specify it to save files), so it doesn't seem that useful. This kind of thing is very useful when you have a long running process (like a web server), as it's a LOT easier to search through a file than a console window. But in this case is doesn't seem to add that much.
https://docs.python.org/3/howto/logging.html is the canonical source of logging info, but is a bit hard to read.

Thomas-Lemoine:
I see. I guess in that case it's more of a print thing. I did set it to be in the log file because I was printing a lot and that file was easier to read but I could just print things while I'm testing it and remove it afterwards

tests/align_data/articles/test_datasets.py
```diff
@@ -356,46 +356,49 @@ def test_arxiv_process_entry(_, mock_arxiv):
...
-        <h2>Download:</h2>
-        <ul><li>Withdrawn</li></ul>
-        <div class="abs-license"><div hidden="">No license for this version due to withdrawn</div></div>
-      </div>
-     </div>
-    """
-
-    with patch('requests.get', return_value=Mock(content=response)):
-        article = dataset.process_entry(item)
-        assert article.status == 'Withdrawn'
```
Thomas-Lemoine, line 381: This kept causing an error, and then I thought it was a merging issue where at some point this was deleted from main but the merge didn't remove it, but I'm not so sure anymore. It appears that article.status is None but article.to_dict["status"] is "Withdrawn"?

align_data/settings.py
```diff
@@ -34,6 +34,7 @@
 port = os.environ.get("ARD_DB_PORT", "3306")
 db_name = os.environ.get("ARD_DB_NAME", "alignment_research_dataset")
 DB_CONNECTION_URI = f"mysql+mysqldb://{user}:{password}@{host}:{port}/{db_name}"
+ARTICLE_MAIN_KEYS = ["id", "source", "title", "authors", "text", "url", "date_published"]
```
mruwnik, line 37: change this to `ARTICLE_MAIN_KEYS = ["id", "source", 'source_type', "title", "authors", "text", "url", "date_published", "status", "comments"]` for your failing test to start working

tests/align_data/articles/test_datasets.py
mruwnik: alignmnet_dataset adds the status to the metadata rather than to the column. You'll need to update the list of columns in settings.py

align_data/sources/arbital/arbital.py
```diff
@@ -1,32 +1,31 @@
...
-    else:
-        title = url
+    text_parts = contents[1].split(" ")
+    url = f"https://arbital.com/p/{text_parts[0]}"
+    title = " ".join(text_parts[1:]) if len(text_parts) > 1 else url
     return f"[{title}]({url})"
 
 
-def flatten(val):
+def flatten(val): # val is Union[List, Tuple, str] and returns List[str]
```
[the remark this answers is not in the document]
Thomas-Lemoine: Yeah I did, mb, I did the same thing you did yesterday night, looking through the changes and seeing what didn't make sense, and I did add that and removed the query_items deplicate, and a few other things. just pushed that right now

align_data/settings.py
Thomas-Lemoine: Ah I see, nice catch

align_data/sources/articles/datasets.py
```diff
@@ -46,7 +46,11 @@ def items_list(self):
         url = f'https://docs.google.com/spreadsheets/d/{self.spreadsheet_id}/export?format=csv&gid={self.sheet_id}'
         logger.info(f'Fetching {url}')
         df = pd.read_csv(url)
-        return (item for item in df.itertuples() if self.get_item_key(item))
+        return (
+            item 
+            for item in df.itertuples() 
+            if not pd.isna(self.get_item_key(item))
```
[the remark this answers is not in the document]
Thomas-Lemoine: ah, self.get_item_key(item) uses self.maybe which uses pd.isna anyway, right? In tests I just ran, they have the same output; seems like self.get_item_key(item) will be None sometimes (ie when url is missing) and in that case it counts as isna. In any case I removed it, it filters on "self.get_item_key(item) is not None" now, which makes more sense

align_data/sources/articles/google_cloud.py
```diff
@@ -231,5 +232,7 @@ def google_doc(url: str) -> Optional[str]:
 
     doc_id = res.group(1)
     body = fetch_element(f'https://docs.google.com/document/d/{doc_id}/export?format=html', 'body')
-    if body:
-        return MarkdownConverter().convert_soup(body).strip()
+    if not body:
```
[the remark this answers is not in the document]
Thomas-Lemoine: good idea

align_data/sources/articles/pdf.py
```diff
@@ -72,21 +72,6 @@ def fetch_pdf(link):
         logger.error('Could not read PDF file: %s', e)
         return {'error': str(e)}
 
-    filenames = [
```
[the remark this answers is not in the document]
Thomas-Lemoine: ah I see, yeah

align_data/sources/articles/google_cloud.py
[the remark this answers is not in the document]
Thomas-Lemoine: mypy complains about this since it's possible for body (Tag | None) to have a Tag that is falsy, in which case google_doc outputs str | None | Tag. I guess I can just ignore mypy though that is pretty minor

align_data/sources/articles/parsers.py
```diff
@@ -101,14 +118,29 @@ def getter(url):
...
 
 
-HTML_PARSERS = {
+HTML_PARSERS: Dict[str, HtmlParserFunc] = {
     "academic.oup.com": element_extractor("#ContentTab"),
     "ai.googleblog.com": element_extractor("div.post-body.entry-content"),
-    "arxiv-vanity.com": parse_vanity,
-    "ar5iv.labs.arxiv.org": parse_vanity,
+    # TODO: arxiv-vanity.com does not output the same type as the other parsers: Dict[str, str] instead of str
+    # ar5iv.labs.arxiv.org too. Are these pdf parsers? not rly, but they don't output the same type as the other html parsers
```
mruwnik, line 141: check https://github.com/StampyAI/alignment-research-dataset/pull/138 for a fix. This sort of evolved over time from a simple `return the contents of this link` to `if basic html then just return the text, otherwise return a dict with additional metadata`. The aforementioned PR should move everything over to dicts. Which means that these parsers can also extract e.g. title, authors etc. if it makes sense for them to do so

align_data/sources/articles/parsers.py
```diff
@@ -251,38 +282,48 @@ def getter(url):
 
 
 def parse_domain(url: str) -> str:
-    return url and urlparse(url).netloc.lstrip('www.')
+    def remove_www(net_loc: str) -> str:
+        return net_loc[4:] if net_loc.startswith("www.") else net_loc
+    return remove_www(urlparse(url).netloc)
```
mruwnik, line 287: why add the extra function here? How is it better than just having the code?

align_data/sources/arxiv_papers/arxiv_papers.py
```diff
@@ -32,28 +34,27 @@ def canonical_url(url: str) -> str:
...
-        f"https://ar5iv.org/abs/{paper_id}",
-    ]:
-        if contents := parse_vanity(link):
-            return contents
+def get_contents(paper_id: str) -> Dict[str, Any]:
+    if contents := parse_vanity(f"https://www.arxiv-vanity.com/papers/{paper_id}"):
+        return contents
+    if contents := parse_vanity(f"https://ar5iv.org/abs/{paper_id}"):
+        return contents
     return fetch_pdf(f"https://arxiv.org/pdf/{paper_id}.pdf")
```
mruwnik, line 42: you could even do 
```
return (
    parse_vanity(f"https://www.arxiv-vanity.com/papers/{paper_id}") or
    parse_vanity(f"https://ar5iv.org/abs/{paper_id}") or
    fetch_pdf(f"https://arxiv.org/pdf/{paper_id}.pdf")
)
```

align_data/sources/arxiv_papers/arxiv_papers.py
```diff
@@ -72,21 +73,23 @@ def add_metadata(data, paper_id):
...
     data = add_metadata({
         "url": canonical_url(url),
         "source_type": paper.get('data_source'),
     }, paper_id)
-    authors = data.get('authors') or paper.get("authors") or []
-    data['authors'] = [str(a).strip() for a in authors]
+
+    authors = data.get('authors') or paper.get("authors")
+    if not authors: data['authors'] = []
+    elif not isinstance(authors, list): data['authors'] = [str(authors).strip()]
```
mruwnik, line 91: [NIT] I'm not a fan of these oneliner clauses

mruwnik (commented):
looking good!

requirements.txt
```diff
@@ -27,6 +27,10 @@ google-auth-oauthlib
 google-auth-httplib2
 google-api-python-client
 gspread
+python-dateutil
+bs4
+pytz
+epub_meta
```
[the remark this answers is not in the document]
Thomas-Lemoine: Oh I see. That's hard to tell, do you know of a way of knowing this in general or do you only know it for those ones in particular

align_data/sources/articles/parsers.py
Thomas-Lemoine: good point, not sure why I did all tha
```py
net_loc = urlparse(url).netloc
return net_loc[4:] if net_loc.startswith("www.") else net_loc
```
makes more sense

align_data/sources/arxiv_papers/arxiv_papers.py
Thomas-Lemoine: nice

mruwnik (approved):

Thomas-Lemoine:
not sure if I'm doing this wrong; I keep fetching origin/main to merge onto this branch and then pushing that merge but it makes the whole commit history very messy, and I have to fix some merge conflicts, which get drowned in the merge and is hard to find afterwards.

align_data/common/alignment_dataset.py
```diff
@@ -189,7 +189,7 @@ def unprocessed_items(self, items=None) -> list | filter:
 
         return items_to_process
 
-    def fetch_entries(self) -> Generator[Article, None, None]:
```
mruwnik: why are you removing this?

Thomas-Lemoine: Not much of a reason but the reason was that the ide could already infer the type signature without it being explicitely written

Thomas-Lemoine: I brought it back

mruwnik (approved):
go go go!

PR #103, opened 2023-07-23
Remove old jsonl flows

mruwnik:


Thomas-Lemoine (dismissed):
The refactor of main.py is great. What will happen with rebuilding now though? Does it default to rebuilding or default to not rebuilding? Also, like the 'all' dataset, but is it worth adding checks in case someone gives all *and* other names (which might break things with repeated entries)?

align_data/common/alignment_dataset.py
```diff
@@ -136,15 +133,12 @@ def commit():
...
-        # make sure the path to the raw data exists
-        self.files_path.mkdir(parents=True, exist_ok=True)
-
         self._outputted_items = self._load_outputted_items()
 
     @property
     def items_list(self):
         """Returns a generator of items to be processed."""
-        return self.files_path.glob(self.glob)
+        return []
```
Aprillion, line 141: is there any benefit to the wrong docstring?
a) variable name says "list", docstring "generator"
b) it's a list
c) it's a `@property`, not a callable function, so "_Returns_" can be a bit misleading 🤷‍♀️
d) even if the function returned a generator _instance_ previously, so the function was a "generator", the value _returned_ by a generator is called an _iterator_, right?

Aprillion (approved):

align_data/common/alignment_dataset.py
mruwnik: no, you're right. This is what happens when docstrings don't keep up with the actual code :/

PR #106, opened 2023-07-24
Restructures align_data by putting sources in sources folder

henri123lemoine:


henri123lemoine:
Also moved over most of the old stampy-chat pinecone-updating code. It doesn't work right now because it's not updated to handle mysql, and the sqlite3 part, as well as the part where it gets the huggingface dataset, have been removed. New PR to fix that coming asap

align_data/pinecone/text_splitter.py
```diff
@@ -0,0 +1,102 @@
+# dataset/text_splitter.py
+
+from typing import List, Callable, Any
+from langchain.text_splitter import TextSplitter
+from nltk.tokenize import sent_tokenize
+
+
+class ParagraphSentenceUnitTextSplitter(TextSplitter):
```
Thomas-Lemoine, line 8: curious what the benefits of this are

align_data/pinecone/update_pinecone.py
```diff
@@ -0,0 +1,190 @@
...
+from align_data.settings import USE_OPENAI_EMBEDDINGS, OPENAI_EMBEDDINGS_MODEL, \
+    OPENAI_EMBEDDINGS_DIMS, OPENAI_EMBEDDINGS_RATE_LIMIT, \
+    SENTENCE_TRANSFORMER_EMBEDDINGS_MODEL, SENTENCE_TRANSFORMER_EMBEDDINGS_DIMS, \
+    CHUNK_SIZE, MAX_NUM_AUTHORS_IN_SIGNATURE, EMBEDDING_LENGTH_BIAS
+
+import logging
+logger = logging.getLogger(__name__)
+
+
+class ARDUpdater:
```
Thomas-Lemoine, line 18: does this only update pinecone? if yes, then it should not be named ARDUpdater imo, too vague, but if no, the file name shouldn't be update_pinecone imo

Thomas-Lemoine (approved):

align_data/pinecone/text_splitter.py
henri123lemoine: It splits paragraphs into sentences, but more rapidly and approximately than other similar methods. It could be changed back to the more precise splitters, I suppose, but I don't think it matters.

align_data/pinecone/update_pinecone.py
henri123lemoine: Agreed. I'll change it for the Pinecone PR

PR #108, opened 2023-07-29
Unify ebooks and transcripts

mruwnik:
* add missing requirements
* fix audio_transcripts + move them to the common gdocs for easier management
* ditto for ebooks (a lot of them were already moved)


align_data/sources/articles/datasets.py
```diff
@@ -24,6 +22,13 @@ class SpreadsheetDataset(AlignmentDataset):
     sheet_id: str
     done_key = "title"
     source_filetype = None
+    batch_size = 1
+
+    @staticmethod
+    def is_val(val):
```
Aprillion, line 28: if it doesn't return bool, does it have to be called `is_...`? can it have some name to describe the purpose?

Aprillion (approved):

align_data/sources/articles/datasets.py
mruwnik: I couldn't come up with a better name, as it's supposed to check whether `val` isn't a not-a-number (which is how pandas handles empty values), but it's used in many places as a `val if is_val(val) else None`

PR #109, opened 2023-07-30
Gdocs metadata in gsheets

mruwnik:
Get gdocs metadata from the Gsheet with clean values


.github/workflows/fetch-weekly.yml
```diff
@@ -16,7 +16,6 @@ jobs:
           - alignment_newsletter
           - arbital
           - arxiv
-          - audio_transcripts
```
ccstan99, line 19: Are you removing this from fetch-weekly because we have implemented processing audio transcripts yet? If that's the reasoning, we can remove nonarxiv_papers and reports also, unless you wanted to deal with formally removing those as a separate PR.

align_data/sources/articles/__init__.py
```diff
@@ -28,4 +28,9 @@
         spreadsheet_id='1l3azVJVukGAvZPgg0GyeqiaQe8bEMZvycBJaA8cRXf4',
         sheet_id='1003473759'
     ),
+    DocArticles(
+        name='gdocs',
+        spreadsheet_id='1l3azVJVukGAvZPgg0GyeqiaQe8bEMZvycBJaA8cRXf4',
```
ccstan99, line 33: This is the gsheet "Items metadata"? We're still using this after moving to MySQL?

align_data/sources/articles/datasets.py
```diff
@@ -122,3 +124,18 @@ def _get_text(self, item):
...
         return vals['text']
+
+
+class DocArticles(SpreadsheetDataset):
+
+    source_filetype = 'docx'
+
+    def _get_text(self, item):
+        pandoc_path = Path('data/raw/pandoc/pandoc/')
+        if pandoc_path.exists():
```
ccstan99, line 135: I just noticed that the 11 Vael Gates interviews in gdocs/docx also have a duplicated MD versions in the [audio transcripts folder](https://drive.google.com/drive/u/0/folders/12FFjcP_60rq3qgq_Wl7HDWIjEA7p_3T_)... So maybe a little housekeeping is needed to decided on the ultimate location? :/

ccstan99 (approved):

.github/workflows/fetch-weekly.yml
mruwnik: no, I moved them over to the markdown tab in the gsheets, to keep them in one place

align_data/sources/articles/__init__.py
mruwnik: for now :/ Or at least as sort of a backup. I'm a bit worried that something will happen to nuke the mysql server and all the data will be lost

align_data/sources/articles/datasets.py
mruwnik: yes!! This code is useful anyway, though

PR #110, opened 2023-08-01
handle youtube

mruwnik:


align_data/sources/youtube/youtube.py
```diff
@@ -0,0 +1,145 @@
...
+
+        return self.make_data_entry({
+            "text": contents,
+            "url": video_url,
+            "title": video['snippet']['title'],
+            "source": self.name,
+            "source_type": "youtube",
+            "date_published": self._get_published_date(video),
+            "authors": self.extract_authors(video),
+        })
```
ccstan99, line 102: Did you find that we were able to get fairly decent metadata this way?

align_data/sources/youtube/youtube.py
```diff
@@ -0,0 +1,145 @@
...
+    def _get_contents(self, video):
+        video_id = self._get_id(video)
+        try:
+            transcript = YouTubeTranscriptApi.list_transcripts(video_id).find_transcript(['en', 'en-GB']).fetch()
+            return '\n'.join([i['text'] for i in transcript])
+        except (NoTranscriptFound, VideoUnavailable):
+            return None
+        except TranscriptsDisabled:
+            logger.error(f'Transcripts disabled for https://www.youtube.com/watch?v={video_id} - skipping')
+            return None
```
ccstan99, line 80: Were there many that had transcriptions disabled or unavailable? Would something like this help? https://huggingface.co/spaces/SteveDigital/free-fast-youtube-url-video-to-text-using-openai-whisper

align_data/sources/youtube/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+    ),
+    YouTubePlaylistDataset(
+        name='ai_alignment_playlist',
+        playlist_ids=[
+            'PLqYmG7hTraZCRwoyGxvQkqVrZgDQi4m-5',
+            'PLqYmG7hTraZBiUr6_Qf8YTS2Oqy3OGZEj',
+            'PLAPVC5uNprwY0q4_nyeeHqIT07wZqwjGO',
+            'PLCRVRLd2RhZTpdUdEzJjo3qhmX3y3skWA',
+            'PLTYHZYmxohXpn5uf8JZ2OouB1PsDJAk-x',
+        ]
```
ccstan99, line 37: Nice! Each channel is a dataset and all the playlists are a dataset? So to add an individual video, we add it to a playlist we're already tracking? Might have quite a few duplicates since all of Rob's videos will also be in a playlist too? And are we skipping "Robert Miles 2" channel since it's only a handful of videos and more off topic?

ccstan99 (approved):
Yay!! I think this is one of the major final missing pieces! Now to clean metadata of special docs and make sure new arxiv papers get scraped.

align_data/sources/youtube/youtube.py
mruwnik: yes. The main missing thing is the authors, as it just does the channel name, but even that should be ok.

align_data/sources/youtube/youtube.py
mruwnik: yes. Or some kind of alternative. Issue added - https://github.com/StampyAI/alignment-research-dataset/issues/112

align_data/sources/youtube/__init__.py
mruwnik: Duplicates are fine - they will be ignored. This is part of the reason why I put all the playlists together :D
That's part of the reason I skipped it - I just wanted the main body of this to be merged, as later additions will be simple

PR #111, opened 2023-08-03
Fix actions

mruwnik:
Get all datasets working with the database:
* alignment_newsletter creates summaries
* reports is removed, as it's now a duplicate of other ones
* actions cleaned up


.github/workflows/fetch-dataset.yml
```diff
@@ -55,7 +54,6 @@ on:
           - qualiacomputing
```
ccstan99, line 54: nonarxiv_paper can be removed along with report. They've both been moved to either xml or pdfs.

qualiacomputing can also be removed since just Murphant helped us identify the relevant pages were added to special doc, just need metadata and get scraped.

.github/workflows/upload-to-huggingface.yml
```diff
@@ -0,0 +1,50 @@
+name: Upload datasets to Huggingface
+on:
+  workflow_dispatch: # allow manual triggering
+  schedule:
+    - cron: "0 3 * * 0"  # Every Sunday at 3 AM
+
```
ccstan99, line 6: Are we pushing to HF? Or is HF pulling live data from MySQL?

README.md
```diff
@@ -20,8 +20,8 @@ The following list of sources may change and items may be renamed:
 - [deepmind_blog](https://deepmindsafetyresearch.medium.com/)
 - [distill](https://distill.pub/)
 - [eaforum](https://forum.effectivealtruism.org/) - selected posts
+- ebooks - books include [Superintelligence](https://www.goodreads.com/book/show/20527133-superintelligence), [Human Compatible](https://www.goodreads.com/book/show/44767248-human-compatible), [Life 3.0](https://www.goodreads.com/book/show/34272565-life-3-0), [The Precipice](https://www.goodreads.com/book/show/50485582-the-precipice), and others
```
ccstan99, line 23: Are the copyright books back in? Maybe don't advertise this until we get permissions straightened out? If pinecone embeddings is now done in ARD instead of stampy-chat, we can just make sure the copyrighted material doesn't go to HF?

ccstan99 (approved):
Looks good! You should be able to add the needed keys too.

align_data/common/alignment_dataset.py
```diff
@@ -196,13 +201,42 @@ def process_entry(self, entry):
...
         try:
             # Totally ignore any timezone info, forcing everything to UTC
             return parse(str(date)).replace(tzinfo=pytz.UTC)
         except ParserError:
             pass
         return None
+
+
+class SummaryDataset(AlignmentDataset):
+
```
ccstan99, line 217: FYI importai and ml_safety_newsletter also function as summaries. It's probably buried in one of the issues somewhere. 

Should we treat arxiv abstracts as a summary too? My thought is this would be the field we search when some asks for "the blog or paper about ..."

.github/workflows/upload-to-huggingface.yml
mruwnik: pushing to HF. The HF dataloader class thingy can pull from SQL, but the datasets on their page are static files which can be used by anyone. This step isn't really needed - it's just a nice bonus thing which might as well be left in

README.md
mruwnik: dunno why this was added here...

align_data/common/alignment_dataset.py
mruwnik: I did an explicit issue for them - https://github.com/StampyAI/alignment-research-dataset/issues/113

mruwnik: abstracts should now be saved as summaries

.github/workflows/fetch-dataset.yml
mruwnik: nonarxiv_paper was going through the xml files. I renamed it to be explicit

mruwnik:
Both the [daily](https://github.com/StampyAI/alignment-research-dataset/actions/runs/5757282057) and [weekly](https://github.com/StampyAI/alignment-research-dataset/actions/runs/5755075470) synchs worked, so now all the datasets should be autoingested :D

align_data/common/alignment_dataset.py
```diff
@@ -80,20 +72,29 @@ def __post_init__(self, data_path=Path(__file__).parent / '../../data/'):
...
-        data['authors'] = ','.join(data.get('authors', []))
-        if summary := ('summary' in data and data.pop('summary')):
-            data['summaries'] = [summary]
-        return Article(
+        article.authors = ','.join(authors)
+        if len(article.authors) > 1024:
+            article.authors = ','.join(article.authors[:1024].split(',')[:-1])
+        return article
+
+    def make_data_entry(self, data, **kwargs) -> Article:
```
Thomas-Lemoine, line 82: This is much better! I would have expected _add_authors to be some Article method like set_authors, but this version has the advantage of being right next to the method that uses it so potentially easier to keep track of

PR #114, opened 2023-08-04
Special docs

mruwnik:
* Imports new items directly from special_docs
* Checks the indecices for new items and adds them if it can (a lot of them aren't handled :/)
* moves a couple of functions around


mruwnik:
ping

align_data/sources/articles/parsers.py
```diff
@@ -33,89 +52,16 @@ def medium_blog(url):
...
-    else:
-        result['error'] = f'unknown content type: {content_type}'
-
-    return result
-
-
 def error(error_msg):
     """Returns a url handler function that just logs the provided `error` string."""
-    def func(url):
+    def func(_url):
```
Thomas-Lemoine, line 64: In my testing this caused the error message to be shown twice, but that might have changed

.github/workflows/push-dataset.yml
```diff
@@ -57,6 +58,7 @@ on:
           - openai.research
           - pdfs
           - rob_miles_ai_safety
+          - special_docs
           - vkrakovna_blog
```
ccstan99, line 62: So, we now have datasets called special_docs, indices, pdfs, xmls, html_articles? It's probably fine now just for our internal use and stampy-chat, but the naming convention probably should be rethought on what/how we expose in the public dataset.

align_data/sources/articles/parsers.py
```diff
@@ -167,6 +95,8 @@ def getter(url):
     "judiciary.senate.gov": error(""),
     "taylorfrancis.com": error("Ebooks are not yet handled"),
     "YouTube.com": error("Youtube videos are not yet handled"),
+    "youtube.com": error("Youtube videos are not yet handled"),
+    "YouTube.be": error("Youtube videos are not yet handled"),
```
ccstan99, line 99: I thought youtube was already handled. Was that for playlists and entire channels but not individual videos?

ccstan99 (approved):
I may not the best person to approve this since I haven't had the chance to sit down to test things out, but this looks important. Thanks!

align_data/sources/articles/parsers.py
mruwnik: precisely. I initially considered reusing that code here, but it's a bit more complicated :/

.github/workflows/push-dataset.yml
mruwnik: yes. This is bad, as you note...

align_data/sources/articles/parsers.py
ccstan99: Maybe it'd be easier to just add these to a playlist then? Can we do that using code or need manual labor?

mruwnik: it's doable by code, but might be a massive pain. I'll add an issue

PR #118, opened 2023-08-08
formatting auto fix

henri123lemoine:
This is not a particularly important change or anything, but it did happen once or twice that I noticed inconsistent formatting, so searched for an auto-formatting tool. I have no strong opinions as to whether this is actually a net positive change.


align_data/db/models.py
```diff
@@ -30,61 +39,73 @@ class Summary(Base):
...
+
     def is_metadata_keys_equal(self, other):
         if not isinstance(other, Article):
-            raise TypeError(f"Expected an instance of Article, got {type(other).__name__}")
+            raise TypeError(
+                f"Expected an instance of Article, got {type(other).__name__}"
+            )
         return not any(
-            getattr(self, key, None) != getattr(other, key, None)  # entry_id is implicitly ignored
+            getattr(self, key, None)
```
Thomas-Lemoine, line 83: This is a bit odd

align_data/db/models.py
```diff
@@ -30,61 +39,73 @@ class Summary(Base):
...
+            != getattr(other, key, None)  # entry_id is implicitly ignored
             for key in PINECONE_METADATA_KEYS
         )
 
     def generate_id_string(self) -> str:
-        return ''.join(str(getattr(self, field)) for field in self.__id_fields).encode("utf-8")
+        return "".join(str(getattr(self, field)) for field in self.__id_fields).encode(
```
Thomas-Lemoine, line 89: This is a bit odd

align_data/db/session.py
```diff
@@ -24,6 +24,4 @@ def stream_pinecone_updates(session, custom_sources: List[str]):
     """Yield Pinecone entries that require an update."""
     yield from session.query(Article).filter(
         Article.pinecone_update_required.is_(True)
-    ).filter(
-        Article.source.in_(custom_sources)  
-    ).yield_per(1000)
\ No newline at end of file
+    ).filter(Article.source.in_(custom_sources)).yield_per(1000)
```
Thomas-Lemoine, line 27: This is a bit odd

align_data/pinecone/pinecone_db_handler.py
```diff
@@ -25,58 +32,60 @@ def __init__(
...
             logger.info(f"{self.index_name}:\n{index_stats_response}")
-            
+
     def upsert_entry(self, entry: Dict, upsert_size=100):
         self.index.upsert(
             vectors=list(
                 zip(
-                    [f"{entry['id']}_{str(i).zfill(6)}" for i in range(len(entry['text_chunks']))], 
-                    entry['embeddings'].tolist(), 
+                    [
```
Thomas-Lemoine, line 54: this is much better

align_data/pinecone/text_splitter.py
```diff
@@ -7,29 +7,33 @@
...
     @param truncate_function: A function that truncates a string to a given unit length.
     """
-    
+
     DEFAULT_MIN_CHUNK_SIZE = 900
     DEFAULT_MAX_CHUNK_SIZE = 1100
     DEFAULT_LENGTH_FUNCTION = lambda string: len(string)
-    DEFAULT_TRUNCATE_FUNCTION = lambda string, length, from_end=False: string[-length:] if from_end else string[:length]
+    DEFAULT_TRUNCATE_FUNCTION = (
+        lambda string, length, from_end=False: string[-length:]
```
Thomas-Lemoine, line 21: This is quite odd, but imo this should be a def function anyway. in some branch I'm writing I had this as 
```py
def default_truncate_function(string: str, length: int, from_end: bool = False) -> str:
    return string[-length:] if from_end else string[:length]
```

and then you pass in `DEFAULT_TRUNCATE_FUNCTION = default_truncate_function`

which is less odd

align_data/pinecone/text_splitter.py
```diff
@@ -43,46 +47,54 @@ def split_text(self, text: str) -> List[str]:
             current_block += "\n\n" + paragraph
             block_length = self._length_function(current_block)
 
-            if block_length > self.max_chunk_size:  # current block is too large, truncate it
-                current_block = self._handle_large_paragraph(current_block, blocks, paragraph)
+            if (
```
Thomas-Lemoine, line 50: seems like when the comment is too long they prefer to separate the conditional from its comment to reduce the length of a line, but this is obvious enough that you could remove the comment, idk

align_data/pinecone/text_splitter.py
```diff
@@ -43,46 +47,54 @@ def split_text(self, text: str) -> List[str]:
...
-        
+
         blocks = self._handle_remaining_text(current_block, blocks)
 
         return [block.strip() for block in blocks]
 
     def _handle_large_paragraph(self, current_block, blocks, paragraph):
         # Undo adding the whole paragraph
-        current_block = current_block[:-(len(paragraph)+2)]  # +2 accounts for "\n\n"
+        current_block = current_block[: -(len(paragraph) + 2)]  # +2 accounts for "\n\n"
```
Thomas-Lemoine, line 68: !!!! blasphemous, a space after :

align_data/pinecone/update_pinecone.py
```diff
@@ -29,54 +36,69 @@ class PineconeEntry(BaseModel):
...
 
 class PineconeUpdater:
     def __init__(
-        self, 
+        self,
         min_chunk_size: int = ParagraphSentenceUnitTextSplitter.DEFAULT_MIN_CHUNK_SIZE,
         max_chunk_size: int = ParagraphSentenceUnitTextSplitter.DEFAULT_MAX_CHUNK_SIZE,
-        length_function: Callable[[str], int] = ParagraphSentenceUnitTextSplitter.DEFAULT_LENGTH_FUNCTION,
-        truncate_function: Callable[[str, int], str] = ParagraphSentenceUnitTextSplitter.DEFAULT_TRUNCATE_FUNCTION,
+        length_function: Callable[
```
Thomas-Lemoine, line 68: This seems worse but afaik the convention is defining the length_function's type at the top of the file like as a TypeVar? But maybe that's weird? GPT says
```py
from typing import TypeVar, Callable

StrToIntFunction = Callable[[str], int]
length_function: StrToIntFunction = etc.
```

align_data/pinecone/update_pinecone.py
```diff
@@ -29,54 +36,69 @@ class PineconeEntry(BaseModel):
...
+            [str], int
+        ] = ParagraphSentenceUnitTextSplitter.DEFAULT_LENGTH_FUNCTION,
+        truncate_function: Callable[
```
Thomas-Lemoine, line 71: see previous comment

align_data/pinecone/update_pinecone.py
```diff
@@ -101,45 +125,54 @@ def process_entries(self, article_stream: Generator[Article, None, None]) -> Gen
...
-        
+            embeddings[i] = bias * np.array(embedding["embedding"])
+
         return embeddings
 
     @staticmethod
     def get_authors_str(authors_lst: List[str]) -> str:
-        if authors_lst == []: return 'n/a'
-        if len(authors_lst) == 1: return authors_lst[0]
+        if authors_lst == []:
```
Thomas-Lemoine, line 171: I prefer the previous way personally but idk

align_data/postprocess/postprocess.py
```diff
@@ -1,59 +1,74 @@
...
                 for obj in reader:
-                    self.all_stats[source_name]['num_entries'] = self.all_stats[source_name].get('num_entries' , 0) + 1
-                    self.all_stats[source_name]['num_tokens'] = self.all_stats[source_name].get('num_tokens' , 0) + len(obj['text'].split())
-                    self.all_stats[source_name]['num_chars'] = self.all_stats[source_name].get('num_chars' , 0) + len(obj['text'])
-                    self.all_stats[source_name]['num_words'] = self.all_stats[source_name].get('num_words' , 0) + len(obj['text'].split())
-                    self.all_stats[source_name]['num_sentences'] = self.all_stats[source_name].get('num_sentences' , 0) + len(obj['text'].split('.'))
-                    self.all_stats[source_name]['num_paragraphs'] = self.all_stats[source_name].get('num_paragraphs' , 0) + len(obj['text'].splitlines())
-        
+                    self.all_stats[source_name]["num_entries"] = (
+                        self.all_stats[source_name].get("num_entries", 0) + 1
```
Thomas-Lemoine, line 34: all of this is too much imo, but couldn't we do 
`source_stats = self.all_stats[source_name]`
and then do
`source_stats["num_entries"] = source_stats.get('num_entries', 0) + 1`
?
But really, this should be a default dict right. oh, can't that just be changed with `self.all_stats =
self.all_stats = {key: default_dict() for key in self.source_list}`?
What gpt says, and it sounds right to me, is 
```py
    def __init__(self):
        # Prepare all_stats as a defaultdict of defaultdicts
        self.all_stats = defaultdict(lambda: defaultdict(int))

    def compute_statistics(self) -> None:
        for source_name, path in tqdm(zip(self.source_list, self.jsonl_list)):
            with jsonlines.open(path) as reader:
                for obj in reader:
                    self.all_stats[source_name]["num_entries"] += 1
                    self.all_stats[source_name]["num_chars"] += len(obj['text'])
```

align_data/settings.py
```diff
@@ -1,30 +1,35 @@
...
 
 ### CODA ###
 CODA_TOKEN = os.environ.get("CODA_TOKEN")
 CODA_DOC_ID = os.environ.get("CODA_DOC_ID", "fau7sl2hmG")
-ON_SITE_TABLE = os.environ.get('CODA_ON_SITE_TABLE', 'table-aOTSHIz_mN')
+ON_SITE_TABLE = os.environ.get("CODA_ON_SITE_TABLE", "table-aOTSHIz_mN")
 
 ### GOOGLE DRIVE ###
-PDFS_FOLDER_ID = os.environ.get('PDF_FOLDER_ID', '1etWiXPRl0QqdgYzivVIj6wCv9xj5VYoN')
+PDFS_FOLDER_ID = os.environ.get("PDF_FOLDER_ID", "1etWiXPRl0QqdgYzivVIj6wCv9xj5VYoN")
```
Thomas-Lemoine, line 12: This should absolutely have both the variable name and the .env key name be the same, like either PDF for both or PDFS for both. I added it to my branch that implements the unimplemented parsers bc I spent a long time looking for the issue before figuring out it was this

align_data/settings.py
```diff
@@ -36,14 +41,20 @@
 OPENAI_EMBEDDINGS_DIMS = 1536
 OPENAI_EMBEDDINGS_RATE_LIMIT = 3500
 
-SENTENCE_TRANSFORMER_EMBEDDINGS_MODEL = "sentence-transformers/multi-qa-mpnet-base-cos-v1"
+SENTENCE_TRANSFORMER_EMBEDDINGS_MODEL = (
```
Thomas-Lemoine, line 44: silly

align_data/sources/articles/articles.py
```diff
@@ -104,31 +111,48 @@ def process_spreadsheets(source_sheet, output_sheets):
...
 
 
 def check_new_articles(source_spreadsheet, source_sheet):
     """Goes through the special indices looking for unseen articles."""
     source_sheet = get_sheet(source_spreadsheet, source_sheet)
-    current = {row.get('title'): row for row in iterate_rows(source_sheet)}
-    seen_urls = {url for item in current.values() for url in [item.get('url'), item.get('source_url')] if url}
+    current = {row.get("title"): row for row in iterate_rows(source_sheet)}
+    seen_urls = {
+        url
```
Thomas-Lemoine, line 125: seen_urls = {
    url
    for item in current.values()
    for key in ("url", "source_url")
    if (url := item.get(key)) is not None
}
🤔

align_data/sources/articles/datasets.py
```diff
@@ -40,9 +39,15 @@ def is_val(val):
 
     @property
     def items_list(self):
-        logger.info(f'Fetching https://docs.google.com/spreadsheets/d/{self.spreadsheet_id}/export?format=CS&gid={self.sheet_id}')
-        df = pd.read_csv(f'https://docs.google.com/spreadsheets/d/{self.spreadsheet_id}/export?format=csv&gid={self.sheet_id}')
-        return (item for item in df.itertuples() if not pd.isna(self.get_item_key(item)))
+        logger.info(
+            f"Fetching https://docs.google.com/spreadsheets/d/{self.spreadsheet_id}/export?format=CS&gid={self.sheet_id}"
```
Thomas-Lemoine, line 43: or url = f'https:etc'
logger.info(f'fetching {url}')
df = pd.read_csv(url)

align_data/sources/articles/indices.py
```diff
@@ -11,184 +11,241 @@
...
 
 def format_far_ai(item):
     return {
-        'title': get_text(item, '.article-title'),
-        'url': f'https://www.safe.ai/research{item.select_one(".article-title a").get("href")}',
-        'source_url': item.select_one('div.btn-links a:-soup-contains("PDF")').get('href'),
-        'authors': ', '.join(i.text for i in item.select('.article-metadata a')),
+        "title": get_text(item, ".article-title"),
+        "url": f'https://www.safe.ai/research{item.select_one(".article-title a").get("href")}',
+        "source_url": item.select_one('div.btn-links a:-soup-contains("PDF")').get(
```
Thomas-Lemoine, line 100: this is odd. according to their convention I would have expected
```py
"source_url": (
    item
    .selecte_one(blabla)
    .get("href")
    ),
```
etc.

align_data/sources/articles/pdf.py
```diff
@@ -88,14 +97,16 @@ def get_arxiv_link(doi):
     if res.status_code != 200:
         return None
 
-    vals = [i for i in response.json().get('values') if i.get('type', '').upper() == 'URL']
+    vals = [
```
Thomas-Lemoine, line 100: how about
```py
val = [
    val
    for val in response.json().get("values")
    if val.get("type", "").upper() == "URL"
]
```
?

align_data/sources/articles/pdf.py
```diff
@@ -170,24 +185,28 @@ def get_first_child(item):
...
-    if date_published := contents.select_one('div.ltx_dates'):
-       date_published = date_published.text.strip('()')
+    if date_published := contents.select_one("div.ltx_dates"):
+        date_published = date_published.text.strip("()")
 
-    text = '\n\n'.join([
-        MarkdownConverter().convert_soup(elem).strip()
-        for elem in contents.select('section.ltx_section')
-    ])
+    text = "\n\n".join(
```
Thomas-Lemoine, line 199: does this need the interior list comprehension?

Thomas-Lemoine: edit: pretty sure not but I mean the difference is negligible, I only mention it because your program makes it into a weird looking thing because of it

align_data/sources/blogs/blogs.py
```diff
@@ -9,82 +9,89 @@
...
+            return item_metadata(paper_link.get("href")).get("text")
 
     def extract_authors(self, article):
-        authors = (
-            article.select_one('div:-soup-contains("Authors") + div .f-body-1') or
-            article.select_one('div:-soup-contains("Acknowledgments") + div .f-body-1')
-        )
+        authors = article.select_one(
+            'div:-soup-contains("Authors") + div .f-body-1'
+        ) or article.select_one('div:-soup-contains("Acknowledgments") + div .f-body-1')
```
Thomas-Lemoine, line 79: This is a bit odd

align_data/sources/youtube/youtube.py
```diff
@@ -113,33 +123,40 @@ def collection_ids(self):
         return [self.channel_id]
 
     def next_page(self, collection_id, next_page_token):
-        return self.youtube.search().list(
-            part='snippet',
-            channelId=collection_id,
-            maxResults=50,
-            pageToken=next_page_token
-        ).execute()
+        return (
```
Thomas-Lemoine, line 126: nice

mruwnik (commented):
What did you use? black seems to be the most popular, from what I recall. I personally am not a big fan of it (it doesn't always stick to PEP8), but at least it's consistent

mruwnik (dismissed):

align_data/pinecone/text_splitter.py
henri123lemoine: Agreed, but doesn't this indicate that the line was already pretty bad. It might make more sense to have the `len(paragraph+2` part be defined beforehand. Or to have `[:2-len(paragraph)]`. Many ways to be cleaner, here.

henri123lemoine: (the above is wrong, I misread. Oops)

align_data/sources/articles/datasets.py
henri123lemoine: Careful! It's not the same url!

align_data/sources/articles/indices.py
henri123lemoine: Their convention, as I understand, is to keep it on one line until it reaches the line length limit, and split then. So unluckily, only the href is caught here.

align_data/sources/articles/pdf.py
henri123lemoine: vals* and yes

align_data/sources/youtube/youtube.py
henri123lemoine: I don't get how that's better, maybe I'm just not used to this format.

Thomas-Lemoine (approved):

align_data/sources/youtube/youtube.py
Thomas-Lemoine: I guess each new line is a thing being done to the result so far, so it's a bit easier to read

PR #120, opened 2023-08-09
Regular pinecone updates

mruwnik:
This will still require the appropriate API keys to be set, but it should run as a second job after fetching datasets

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,9 +146,34 @@ There are Datasets defined for various types of data sources - first check if an
 
 When wishing to update the whole dataset, run `python main.py fetch_all`. You can also fetch a specific subsection of a dataset by its name, for example `python main.py fetch aisafety.info`.
 
+## Configuration
+
+Various subcomponents use various external services, so need credentials set. This is done via environment variables, the easiest way of setting which is by copying `~/.env.example` to `~/.env` and changing the appropriate values.
+
+### Logging
+
+The log level can be configured with the `LOG_LEVEL` environment variable. The default level is 'WARNING'.
+
 ### Coda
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -146,10 +146,35 @@ There are Datasets defined for various types of data sources - first check if an
...
+database can be started in Docker by running
+
+    ./local_db.sh
+
+### Pinecone
+
+For Pinecone updates to work, you'll need to configure the API key:
+
+1. Get an API key, as described [here](https://docs.pinecone.io/docs/quickstart#2-get-and-verify-your-pinecone-api-key)
+2. Create a Pinecone index named "stampy-chat-ard" (or whatever is set as `PINECONE_INDEX_NAME`) with the `dotproduct` metric and 1536 dimensions
```
henri123lemoine, line 174:
```py
from align_data.pinecone.pinecone_db_handler import PineconeDB
pinecone_db = PineconeDB(create_index=True)
```
^^ would create the pinecone db for the user if they set up all the env and settings variables correctly. Might be worth mentioning. Also, it might be worth adding as a method in `main.py`, for the user to run if they don't have a pinecone db set up yet that runs the above code. Asking the user to create the Pinecone index by themselves would be more error prone, for instance it should be clear that the index should be created with `metadata_config={"indexed": [our metadata keys]}` as well as the name/dim/metric.

henri123lemoine (approved):

README.md
mruwnik: I added an issue: https://github.com/StampyAI/alignment-research-dataset/issues/121

PR #124, opened 2023-08-11
Special docs

mruwnik:
* Uploads all special docs things as a single item to HF (i.e. 'pdfs', 'xmls', etc.)
* Adds a `status` column - this means that special docs items only need to be imported once, at the cost of having partially filled items in the database
* Items with missing required fields will no longer raise an error - now they'll set the appropriate value in the status field

### DIFF
--- a/align_data/common/alignment_dataset.py
+++ b/align_data/common/alignment_dataset.py
@@ -106,10 +106,14 @@ class AlignmentDataset:
                 jsonl_writer.write(article.to_dict())
         return filename.resolve()
 
+    @property
+    def _query_items(self):
+        return select(Article).where(Article.source == self.name)
+
     def read_entries(self, sort_by=None):
         """Iterate through all the saved entries."""
         with make_session() as session:
-            query = select(Article).where(Article.source == self.name)
+            query = self._query_items
             if sort_by is not None:
                 query = query.order_by(sort_by)
             for item in session.scalars(query):
--- a/align_data/common/alignment_dataset.py
+++ b/align_data/common/alignment_dataset.py
@@ -85,7 +85,7 @@ class AlignmentDataset:
 
         article = Article(
             id_fields=self.id_fields,
-            meta={k: v for k, v in data.items() if k not in INIT_DICT},
+            meta={k: v for k, v in data.items() if k not in INIT_DICT and v is not None},
             **{k: v for k, v in data.items() if k in INIT_DICT},
         )
         self._add_authors(article, authors)
--- a/align_data/db/models.py
+++ b/align_data/db/models.py
@@ -58,6 +58,7 @@ class Article(Base):
         DateTime, onupdate=func.current_timestamp()
     )
     status: Mapped[Optional[str]] = mapped_column(String(256))
+    comments: Mapped[Optional[str]] = mapped_column(LONGTEXT)  # Editor comments. Can be anything
 
     pinecone_update_required: Mapped[bool] = mapped_column(Boolean, default=False)
 
--- a/align_data/db/models.py
+++ b/align_data/db/models.py
@@ -11,11 +11,10 @@ from sqlalchemy import (
     String,
     Boolean,
     Text,
-    Float,
     func,
     event,
 )
-from sqlalchemy.orm import DeclarativeBase, Mapped, Session, mapped_column, relationship
+from sqlalchemy.orm import DeclarativeBase, Mapped, mapped_column, relationship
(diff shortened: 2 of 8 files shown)


align_data/db/models.py
```diff
@@ -120,12 +124,18 @@ def _set_id(self):
...
 
         if target.id:
             target.verify_id()
         else:
             target._set_id()
+
+        # This assumes that status pretty much just notes down that an entry is invalid. If it has
+        # all fields set and is being written to the database, then it must have been modified, ergo
+        # should be also updated in pinecone
+        if not target.status:
```
mruwnik, line 138: I'm not sure about this. Practically speaking, the column should probably be 'errors' or something. But this seems more generic?

align_data/common/alignment_dataset.py
```diff
@@ -85,7 +85,7 @@ def make_data_entry(self, data, **kwargs) -> Article:
 
         article = Article(
             id_fields=self.id_fields,
-            meta={k: v for k, v in data.items() if k not in INIT_DICT},
+            meta={k: v for k, v in data.items() if k not in INIT_DICT and v is not None},
```
mruwnik, line 88: no point in the metadata dict having `None` values, so strip 'em out

align_data/db/models.py
ccstan99: 'status' works for me. Maybe add string 'comments' to elaborate on status or confidence score.

align_data/sources/articles/datasets.py
```diff
@@ -74,24 +76,25 @@ def process_entry(self, item):
 
 class SpecialDocs(SpreadsheetDataset):
 
+    @property
+    def _query_items(self):
+        special_docs_types = ["pdf", "html", "xml", "markdown", "docx"]
```
ccstan99, line 81: Need to be more consistent in source field values. We currently have: html and html_articles, pdf and pdfs, xml and maybe more.

ccstan99 (approved):
General issues from our earlier conversation:
1. In additional to 'status' column, also add 'comments' which is a string to elaborate on status or confidence score
2. Fix a number of difference sources from special docs should all go into a "misc" HF bucket: html/html_articles, pdf/pdfs, xml
3. Will write a separate issue to handle arxiv papers from special docs

align_data/sources/articles/datasets.py
mruwnik: I'll run a manual SQL query once this is merged to clean things up. I considered adding it in a migration, but it doesn't seem worth adding cruft

align_data/sources/articles/datasets.py
```diff
@@ -74,24 +76,25 @@ def process_entry(self, item):
...
+
     def process_entry(self, item):
         metadata = {}
         if url := self.maybe(item.source_url) or self.maybe(item.url):
             metadata = item_metadata(url)
 
-        text = metadata.get('text')
-        if not text:
-            logger.error('Could not get text for %s - skipping for now', item.title)
-            return None
```
mruwnik, line 85: this means that pretty much everything is added to the database

align_data/sources/articles/datasets.py
ccstan99: If it's just a 1-time fix then probably no need for a migration?

align_data/sources/articles/datasets.py
ccstan99: I just started looking at entries that are being pulled in and some don't look quite right. I'm slightly hesitant about dumping everything into the DB since it looks like we picked up quite a few invalid urls. Let me see if I can figure out a pattern.

ccstan99 (approved):

align_data/sources/articles/datasets.py
mruwnik: this is an issue. Both with them being saved to the database and them gumming up the works. It would be nice to have an extra "check if this item looks at all decent" function to filter these out. As in any item that starts with `hhtps://` won't work for sure, same if it doesn't have an url etc.
There probably won't be a good solution. One the one hand you want to import as much as possible, so the DB really is a valid source of truth and the special docs can be changed as much as desired. on the other hand importing bad data is a waste of time and energy so should be avoided... Why is life so hard?! :D

ccstan99: "check if this item looks at all decent" seems like a good idea. Part of our challenge is wanting to automatically update using indices which could be faulty. Maybe it's ok we just miss some "good" data or just log things we need to consider?

PR #129, opened 2023-08-12
Allow manual batch metadata updates

mruwnik:
* Adds a command to batch update metadata from a csv file (`python main.py update <path to file>`) which can be used on the export of phpmyadmin, or google sheets
* Adds a github action that receives an url to a csv and applies it to the above command

So the updating flow can be:
1. give someone a csv dump of all rows that need to be updated
2. they update it e.g. in google sheets
3. once they've cleaned things up, the url of the google sheet is provided to the github action, which will then import all the changes
4. profit

Or they could host the csv somewhere like s3 and provide it from there, as it doesn't matter to the script - as long as it can download the file as a csv, it's happy


align_data/sources/articles/updater.py
```diff
@@ -0,0 +1,87 @@
...
+        if pd.isna(val):
+            return None
+        return val
+
+    @property
+    def items_list(self):
+        df = pd.read_csv(self.csv_path, delimiter=self.delimiter)
+        self.csv_items = [
+            item for item in df.itertuples()
+            if self.maybe(item, 'id') or self.maybe(item, 'hash_id')
```
mruwnik, line 37: this allows items to be identified either by internal id, or by pinecone id. Internal ids are nice in that they uniquely identify an item, even if the url or title are changed (and so the hash_id). Hash_id is nice, as it can be applied to any database, and should work for everyone.

align_data/sources/articles/updater.py
```diff
@@ -0,0 +1,87 @@
...
+        # If no url found, then don't bother fetching the text - assume it was successfully fetched previously
+        url = self.maybe(updates, 'source_url') or self.maybe(updates, 'url')
+        if not url:
+            return
+
+        metadata = item_metadata(url)
+        # Only change the text if it could be fetched - better to have outdated values than none
+        if metadata.get('text'):
+            article.text = metadata.get('text')
+        article.status = metadata.get('error')
```
mruwnik, line 68: a nice side effect of this is that it will clear out any previous errors if the text could successfully be found

align_data/sources/articles/updater.py
```diff
@@ -0,0 +1,87 @@
...
+        return select(Article).where(or_(Article.id.in_(hash_ids), Article._id.in_(ids)))
+
+    def update_text(self, updates, article):
+        # If the url is the same as it was before, and there isn't a source url provided, assume that the
+        # previous text is still valid
+        if article.url == self.maybe(updates, 'url') and not self.maybe(updates, 'source_url'):
+            return
+
+        # If no url found, then don't bother fetching the text - assume it was successfully fetched previously
+        url = self.maybe(updates, 'source_url') or self.maybe(updates, 'url')
```
mruwnik, line 60: this allows for replacing the text with contents from a different url than is displayed to the user - quite useful e.g. in the case of ebooks, where you want them to have a nice url to goodreads (or whatever), but also want to be able to fetch the contents

main.py
```diff
@@ -44,20 +44,6 @@ def fetch(self, *names) -> None:
 
             dataset.add_entries(dataset.fetch_entries())
 
-    def fetch_all(self, *skip) -> None:
```
mruwnik, line 47: `python main.py fetch_all` == `python main.py fetch all`, so removing this means there is less to maintain. Though it's handy to be able to say "get all items apart from ..." rather than having to explictly list them out, so it might be better to leave it?

align_data/sources/articles/updater.py
ccstan99: Nice!

align_data/sources/articles/updater.py
ccstan99: Are we storing the source_url in a column in the DB or just checking as we read the metadata? If we're not storing it, should we for reproducibility sake?

main.py
ccstan99: Not sure I have a strong opinion on this one. Shall we discuss at the meeting?

ccstan99 (dismissed):

align_data/sources/articles/updater.py
mruwnik: I'll add it to the metadata. I was initially worried about leaking access to the actual source, but the contents are already in the text column, so that doesn't really matter all that much...

ccstan99 (approved):

Issue #122, opened 2023-08-10
Fix NULL authors


ccstan99:
Some greaterwrong (lesswrong, eaforum, alignmentforum) posts have anonymous authors. We should list those authors as "anonymous" instead of NULL in the database, which causes them to be considered "incomplete"
Examples: [[1]](https://www.lesswrong.com/posts/QTjrbk43HD4qevFEk/conditional-on-the-first-agi-being-aligned-correctly-is-a) [[2]](https://forum.effectivealtruism.org/posts/D8NfyuQeGspM9fYpT/begging-pleading-ai-orgs-to-comment-on-nist-ai-risk) [[3]](https://www.alignmentforum.org/posts/KSWSkxXJqWGd5jYLB/the-speed-simplicity-prior-is-probably-anti-deceptive)

The default author for `openai.research` dataset should be "OpenAI Research" instead of NULL.
The default author for `distill` dataset should be "Distill" instead of NULL.

PR #130, opened 2023-08-12
Default authors

mruwnik:
https://github.com/StampyAI/alignment-research-dataset/issues/122


ccstan99 (approved):
Looks good! I suspect we'll see more when we start looking through all the html in special_docs. Maybe a good default in general might be the domain, i.e. safe.ai, anthropic.com, longtermrisk.org, futureoflife.org (but NOT drive.google.com) without hand-coding all possibilities?

mruwnik:
Maybe... But that would probably have to be sort of a white list

Issue #125, opened 2023-08-11
Properly handle arxiv papers


ccstan99:
- Make sure arxiv papers (anything in arxiv domain) suggested or found from indices are handled properly by arxiv parser and not treated as a regular pdf

- For url consistency, use the canonical format:
https://arxiv.org/abs/####.####
https://arxiv.org/abs/2110.08058 instead of https://arxiv.org/pdf/2110.08058.pdf
https://arxiv.org/abs/math/0006233 instead of https://arxiv.org/pdf/math/0006233v2.pdf

- Need to track & update for multiple/latest versions of paper
https://arxiv.org/abs/1606.08514v2 has latest version 4

- Check status for withdrawn papers, for example:
https://arxiv.org/abs/1805.08180
https://arxiv.org/abs/1805.08347
https://arxiv.org/abs/2006.06870

PR #131, opened 2023-08-12
Handle Arxiv links in special docs

mruwnik:
https://github.com/StampyAI/alignment-research-dataset/issues/125

* arxiv articles in special_docs are handled with the arxiv parser code
* Saves the canonical url
* Checks if articles are retracted

### DIFF
--- a/align_data/sources/articles/datasets.py
+++ b/align_data/sources/articles/datasets.py
@@ -2,6 +2,7 @@ import logging
 import os
 from dataclasses import dataclass
 from pathlib import Path
+from typing import Dict
 from urllib.parse import urlparse
 
 import pandas as pd
@@ -13,8 +14,11 @@ from sqlalchemy import select
 from align_data.common.alignment_dataset import AlignmentDataset
 from align_data.db.models import Article
 from align_data.sources.articles.google_cloud import fetch_file, fetch_markdown
-from align_data.sources.articles.parsers import HTML_PARSERS, extract_gdrive_contents, item_metadata
+from align_data.sources.articles.parsers import (
+    HTML_PARSERS, extract_gdrive_contents, item_metadata, parse_domain
+)
 from align_data.sources.articles.pdf import read_pdf
+from align_data.sources.arxiv_papers.arxiv_papers import fetch as fetch_arxiv
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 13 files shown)


align_data/sources/articles/datasets.py
```diff
@@ -172,3 +178,26 @@ def _get_text(self, item):
         file_id = item.source_url.split("/")[-2]
         file_name = fetch_file(file_id)
         return convert_file(file_name, "md", format="docx", extra_args=["--wrap=none"])
+
+
+class ArxivPapers(SpreadsheetDataset):
```
mruwnik, line 183: this got moved over to avoid circular dependancies

align_data/sources/arxiv_papers/arxiv_papers.py
```diff
@@ -1,75 +1,86 @@
...
+
+def get_id(url: str) -> Optional[str]:
+    if res := re.search(r"https?://arxiv.org/(?:abs|pdf)/(.*?)(?:v\d+)?(?:/|\.pdf)?$", url):
+        return res.group(1)
+
+
+def canonical_url(url: str) -> str:
+    if paper_id := get_id(url):
+        return f'https://arxiv.org/abs/{paper_id}'
+    return url
```
ccstan99, line 32: We have a few entries that don't quite fit. Can they be accommodated easily?
https://arxiv.org/abs/astro-ph/0512204
https://arxiv.org/abs/cond-mat/0110359
https://arxiv.org/abs/math/0006233

align_data/sources/arxiv_papers/arxiv_papers.py
```diff
@@ -1,75 +1,86 @@
...
+
+def fetch(url) -> Dict:
+    paper_id = get_id(url)
+    if not paper_id:
+        return {'error': 'Could not extract arxiv id'}
+
+    metadata = get_arxiv_metadata(paper_id)
+
+    if check_if_retracted(url):
+        paper = {'status': 'Retracted'}
```
ccstan99, line 64: Nitpick. Slight preference for keeping the word 'Withdrawn' as status just to match arxiv's status unless you have a specific reason to differ? And for simplicity, since it returns a boolean, just call the function `withdrawn`, i.e.
`if withdrawn(url):`

ccstan99 (approved):
I see you're storing the version. We'll work on checking for updates later?

And are we still using the arxiv spreadsheet tab or will future entries eventually be added directly via suggestions and indices?

align_data/sources/arxiv_papers/arxiv_papers.py
mruwnik: that's fine - the id for these are `astro-ph/0512204` etc. That being said, this will only work for urls that start with `https://arxiv.org/abs/...` or `https://arxiv.org/pdf/....`, so if some new format is introduced, that might cause problems

align_data/sources/arxiv_papers/arxiv_papers.py
mruwnik: I'll change it to `is_withdrawn` (this is yet another place where I miss lisp - there it'd simply be `withdrawn?`...)

mruwnik:
checking for updates can be done later, with the general url checker thingy. Future entries can be added either via suggestions, or directly to the arxiv tab. In the later case it's important that all the metadata is set and correct. So it's probably best for new items to go to suggestions, and the tab to be repopulated every now and then with a dump from the database. Or it can be removed all together, though it's potentially useful as a backup

align_data/sources/arxiv_papers/arxiv_papers.py
ccstan99: This should be fine to handle most cases for now. I think I saw something for psyarxiv or something but those should be rare.

align_data/sources/arxiv_papers/arxiv_papers.py
ccstan99: Lisp! OMG, haven't worked with that in decades.

ccstan99 (dismissed):
Maybe special docs or suggestions can work like the metadata update and just accept a csv url. The indices need to be checked periodically to automatically get newly released papers.

PR #137, opened 2023-08-15
Various minor bug fixes

mruwnik:
* Ignore items that are missing id_fields
* move id_fields to the Article model
* add a magic `is_valid` property to the Article model that can be used both in python and SQL
* make special docs check more possibilities of urls when filtering out seen items 
* make sure all prerequirements for tokenization are available
* save to pinecone in batches - if a batch fails, ignore it. Future runs can add it


align_data/db/models.py
```diff
@@ -86,11 +82,19 @@ def is_metadata_keys_equal(self, other):
...
         )
 
-    def generate_id_string(self) -> str:
+    def generate_id_string(self) -> bytes:
         return "".join(str(getattr(self, field)) for field in self.__id_fields).encode(
             "utf-8"
         )
 
+    @property
+    def __id_fields(self):
```
mruwnik, line 91: the problem with only specifying the id fields in the datasources is that there's no way to know which columns to use when reading them back in. So queries like `session.query(Article).all()` will fetch all articles, but then there's no decent way to check whether the hash_id is correct. Moving it in here and having it depend on the source column removes that issue, at the cost of making it a bit harder to understand the codebase

align_data/db/session.py
```diff
@@ -23,5 +23,9 @@ def make_session(auto_commit=False):
 def stream_pinecone_updates(session, custom_sources: List[str]):
     """Yield Pinecone entries that require an update."""
     yield from session.query(Article).filter(
-        Article.pinecone_update_required.is_(True)
-    ).filter(Article.source.in_(custom_sources)).yield_per(1000)
+        Article.pinecone_update_required.is_(True),
+    ).filter(
+        Article.is_valid
```
mruwnik, line 28: don't embed invalid items (without text, url or title, or with status set to and invalid state)

mruwnik:
things have been inserted to mysql, other things are being added to pinecone: https://github.com/StampyAI/alignment-research-dataset/actions/runs/5878178080/job/15939817629
I don't know how decent they are, only that they're being inserted, which is the main thing now, maybe? hopefully? potentially?

align_data/common/alignment_dataset.py
```diff
@@ -87,7 +84,7 @@ def make_data_entry(self, data, **kwargs) -> Article:
         authors = data.pop("authors", [])
 
         article = Article(
-            id_fields=self.id_fields,
```
Thomas-Lemoine, line 90: how is id_fields used by article if it isn't passed this way?

align_data/db/models.py
```diff
@@ -66,12 +68,6 @@ class Article(Base):
         back_populates="article", cascade="all, delete-orphan"
     )
 
-    __id_fields = ["url", "title"]
-
-    def __init__(self, *args, id_fields, **kwargs):
-        self.__id_fields = id_fields
```
Thomas-Lemoine, line 72: This means the default ['url', 'title'] is only set within AlignmentDataset, right? if so, my previous question applies doubly

align_data/db/models.py
Thomas-Lemoine: what do you mean by reading them back in? I don't understand the specific problem but the trade-off makes sense and this seems fine, we'll have to remove id_fields from __init.py__ files and as fields used within Dataset classes and all that

Ignore my earlier comments though, this makes sense now as a functioning replacement.

align_data/db/models.py
```diff
@@ -129,6 +133,25 @@ def add_meta(self, key, val):
             self.meta = {}
         self.meta[key] = val
 
+    @hybrid_property
+    def is_valid(self):
+        return (
```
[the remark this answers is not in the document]
Thomas-Lemoine: nice

align_data/db/session.py
Thomas-Lemoine: nice. do you have a preference between these? I intuitively like this format (see below) better but, if it's a less conventional format, it might be worse to use in general. I'm asking mostly for the sake of knowing what style to prioritize in the future, this specific example doesn't matter much obv.
```py
    yield from (
        session
        .query(Article)
        .filter(Article.pinecone_update_required.is_(True))
        .filter(Article.is_valid)
        .filter(Article.source.in_(custom_sources))
        .yield_per(1000)
        )
```

align_data/db/models.py
mruwnik: say you have 3 articles (assume all other fields are set):
```
a1 = Article(id_keys=['url', 'title'])
a2 = Article(id_keys=['title', 'url'])
a3 = Article(id_keys=['url'])
a4 = Article(id_keys=['title'])
```

you add them to the database, so they all have hash_id set. Now you fetch them and check they're correct:

```
a1, a2, a3, a4 = session.query(Article).all()
a1.verify_id()  # this one is fine, as it uses the default __id_fields
# these following ones will raise assertion errors, as their hash_ids are checked using the defailt __id_fields
# of ['url', 'title']
a2.verify_id()
a3.verify_id()
a4.verify_id()    
```

This will cause the pinecone update to fail, as part of it is setting `pinecone_update_required = True`, which will raise an exception once it runs `verify_id()`. Strictly speaking, this is only a problem for datasets that don't use the default id_keys, but you get the general issue

align_data/db/session.py
mruwnik: No, either is fine. As long as they're consistently used throughout the codebase. I can change it to this

align_data/db/models.py
Thomas-Lemoine: Hmmm I see, so .verify_id() does not do what I expected it to do, and so presumably generate_id_string doesn't either. Is the issue that the article objects you query from the session don't have access to their own __id_fields ?

mruwnik: pretty much. __id_fields isn't saved to the database. So when you query an article in the database, you get the id hash, but don't know what it was made from. This is because id_fields is controlled by whoever constructs the Article instance, rather than on the basis of whatever is to be saved to the database.
`verify_id()` and `generate_id_string()` work fine - they do what they're supposed to do. But only if you know exactly which id_fields to use. This is fine when you're initially saving items to the database, as they're done by the appropriate dataset, which explicitly sets it for them. But when you're reading them back in, they're all generic Articles (rather than less wrong articles, stampy articles etc.). So you have to work out what id_fields they should have, which can only be done on the basis of the `source` column, as there is no extra info in the database.
An alternative would be to explicitly save the __id_info to the database (e.g. in the `metadata` object), but that seems excessive.

mruwnik:
ping

ccstan99 (dismissed):

PR #138, opened 2023-08-15
Special indices should fetch arxiv papers

mruwnik:
https://github.com/StampyAI/alignment-research-dataset/issues/135

* Make sure all parsers return dicts with metadata
* tidy up arxiv parser
* special indices only adds arxiv papers as proper entries - other ones are marked as ignored so future runs skip them


align_data/sources/articles/datasets.py
```diff
@@ -91,22 +91,22 @@ def _query_items(self):
...
 
     def get_contents(self, item) -> Dict:
-        metadata = {}
+        contents = {}
         if url := self.maybe(item, "source_url") or self.maybe(item, "url"):
-            metadata = item_metadata(url)
+            contents = item_metadata(url)
 
-        return {
+        return dict(contents, **{
```
Thomas-Lemoine, line 98: nice

align_data/sources/articles/indices.py
```diff
@@ -276,16 +276,25 @@ def process_entry(self, item):
...
+
         return self.make_data_entry({
             'source': self.name,
             'url': self.get_item_key(item),
             'title': item.get('title'),
             'date_published': self._get_published_date(item.get('date_published')),
             'authors': self.extract_authors(item),
-            'text': text,
-        })
+        }, **metadata)
```
Thomas-Lemoine, line 300: ah, the metadata contains the text here, right? This is nice but also makes it harder to know the text is in there (where knowing that the text is given to make_data_entry is important). not sure if that matters enough to change anything though

align_data/sources/articles/indices.py
```diff
@@ -276,16 +276,25 @@ def process_entry(self, item):
         if url := item.get('source_url') or item.get('url'):
             metadata = item_metadata(url)
 
-        text = metadata.get('text')
-        if not text:
+        if not metadata.get('text'):
             logger.error('Could not get text for %s (%s) - %s - skipping for now', item.get('title'), url, metadata.get('error'))
             return None
 
+        if parse_domain(metadata.get('source_url') or '') != 'arxiv.org':
```
Thomas-Lemoine, line 283: metadata.get('source_url', '') :p 
I notice: when domain is arxiv.org, different `status, comment, and no metadata or text`. when domain is not arxiv.org, we output the whole metadata including text instead of text only. 

I don't see specific problems, but I don't understand the purpose of this exactly

align_data/sources/articles/parsers.py
```diff
@@ -38,33 +40,47 @@ def getter(url: str):
...
-    return MarkdownConverter().convert_soup(article).strip()
+    source_type = "MediumParser(name='html', url='')"
+    ignored_selectors = ["div:first-child span"]
+
+    def _get_published_date(self, contents):
+        possible_date_elements = contents.select("article div:first-child span")
+        return self._find_date(possible_date_elements)
+
+    def __call__(self, url):
+        return self.get_contents(url)
```
Thomas-Lemoine, line 75: This essentially makes it so 
```py
"deepmindsafetyresearch.medium.com": MediumParser(name='html', url='deepmindsafetyresearch.medium.com'),
```
outputs an object, a MediumParser, but if you run this object, with obj(url), it outputs the contents at that url, found using HTMLDataset's get_contents method, iiuc. Nice!

align_data/sources/articles/parsers.py
```diff
@@ -267,17 +283,18 @@ def item_metadata(url) -> Dict[str, str]:
         # If the url points to a html webpage, then it either contains the text as html, or
         # there is a link to a pdf on it
         if parser := HTML_PARSERS.get(domain):
-            if res := parser(url):
+            res = parser(url)
+            if res and 'error' not in res:
                 # Proper contents were found on the page, so use them
-                return {'source_url': url, 'source_type': 'html', 'text': res}
+                return res
```
Thomas-Lemoine, line 289: Nice

align_data/sources/articles/pdf.py
```diff
@@ -105,10 +105,6 @@ def get_arxiv_link(doi):
     return vals[0]["data"]["value"].replace("/abs/", "/pdf/") + ".pdf"
 
 
-def get_arxiv_pdf(link):
```
Thomas-Lemoine, line 108: this breaks line 115 for me, see get_arxiv_link later in the file

tests/align_data/articles/test_datasets.py
```diff
@@ -51,7 +51,7 @@ def mock_arxiv():
     arxiv = Mock()
     arxiv.Search.return_value.results.return_value = iter([metadata])
 
-    with patch("align_data.sources.arxiv_papers.arxiv_papers.arxiv", arxiv):
```
Thomas-Lemoine, line 54: Is this on purpose? It appears that, in sources/, there's:
```
align_data/sources
├── alignment_newsletter
│   ├── __init__.py
│   └── alignment_newsletter.py
├── arbital
│   ├── __init__.py
│   └── arbital.py
├── articles
│   ├── __init__.py
│   ├── articles.py
│   ├── datasets.py
│   ├── google_cloud.py
│   ├── html.py
│   ├── indices.py
│   ├── parsers.py
│   ├── pdf.py
│   └── updater.py
├── arxiv_papers.py
├── blogs
│   ├── __init__.py
│   ├── blogs.py
│   ├── gwern_blog.py
│   ├── medium_blog.py
│   ├── substack_blog.py
│   └── wp_blog.py
├── distill
│   ├── __init__.py
│   └── distill.py
├── ebooks
│   ├── __init__.py
│   └── agentmodels.py
├── greaterwrong
│   ├── __init__.py
│   └── greaterwrong.py
├── reports
├── stampy
│   ├── __init__.py
│   └── stampy.py
└── youtube
    ├── __init__.py
    └── youtube.py
```
So arxiv_papers.py is the only one that is a file by itself. Would it not make more sense as a part of articles? The only places where I see it currently imported are from article/ files. I may be confused about this

align_data/sources/articles/pdf.py
Thomas-Lemoine: I can't quite create a comment on it, but later in this file there's parse_vanity; it's the only HtmlParserFunc that has type 
`Callable[[str], Dict[str, Any] | None]` instead of
`Callable[[str], Dict[str, Any]]`
which is caused by 
    if not contents:
        return None
where we could simply return {'error': 'no contents found in vanity url'} or something, I think? Would that make more sense

Thomas-Lemoine: ^ to my previous comment, I notice that no, that would break something else, ie 
```py
def get_contents(paper_id: str) -> Dict[str, Any]:
    return (
        parse_vanity(f"https://www.arxiv-vanity.com/papers/{paper_id}") or
        parse_vanity(f"https://ar5iv.org/abs/{paper_id}") or
        fetch_pdf(f"https://arxiv.org/pdf/{paper_id}.pdf")
    )
```

Thomas-Lemoine:
```py
def get_contents(paper_id: str) -> Dict[str, Any]:
    arxiv_vanity = parse_vanity(f"https://www.arxiv-vanity.com/papers/{paper_id}")
    if 'error' not in arxiv_vanity:
        return arxiv_vanity
    
    ar5iv = parse_vanity(f"https://ar5iv.org/abs/{paper_id}")
    if 'error' not in ar5iv:
        return ar5iv
    
    return fetch_pdf(f"https://arxiv.org/pdf/{paper_id}.pdf")
```

align_data/sources/articles/datasets.py
```diff
@@ -18,7 +18,7 @@
     HTML_PARSERS, extract_gdrive_contents, item_metadata, parse_domain
 )
 from align_data.sources.articles.pdf import read_pdf
-from align_data.sources.arxiv_papers.arxiv_papers import fetch as fetch_arxiv
```
Thomas-Lemoine, line 21: how about this becomes
`from align_data.sources.articles.arxiv_papers import fetch_arxiv`,
fetch becomes fetch_arxiv (more descriptive name to avoid mixing them up elsewhere)? would anything else break / is there a disadvantage to that

Thomas-Lemoine (dismissed):

align_data/sources/articles/datasets.py
mruwnik: mainly that a `fetch` function in the `arxiv_papers` module is obviously going to fetch arxiv stuff, so no point in adding the name. Which as you rightly note, is annoying when importing, as then you want it as `arxiv_fetch` so as to not mix it up with other things :D 
I wasn't sure about this - will change it to `fetch_arxiv`

align_data/sources/articles/html.py
```diff
@@ -64,11 +64,15 @@ def element_extractor(selector, remove=[]):
...
         elem = fetch_element(url, selector)
         if not elem:
-            return None
+            return {}
 
         for sel in remove:
             for e in elem.select(sel):
                 e.extract()
-        return MarkdownConverter().convert_soup(elem).strip()
+        return {
```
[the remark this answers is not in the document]
mruwnik: yes. That should be true of all parsers now

align_data/sources/articles/indices.py
mruwnik:
```
{}.get('source_url') == {}.get('source_url', None) == None
{'source_url': 'bla'}.get('source_url') == {'source_url': 'bla'}.get('source_url', None) == 'bla'

({'source_url': None}.get('source_url') or '') == ''
{'source_url': None}.get('source_url', '') == None
```
This will ensure that `parse_domain` always gets a string (unless the `source_url` is really bad, in which case there are bigger problems...)

when the item is an arxiv article, then import it. If it's not, then it should be ignored. But if this simply returns None, then the paper will be retried everytime this is run. If there are a lot of non arxiv things, this can start to add up. So just add it to the database with an `Ignored` status to mark it as seen, but discarded for now - if the editors decide it's worth having, it can be imported then. 

I don't like this. I don't like it at all. But I couldn't come up with any simpler way to mark items that should be ignored on subsequent runs

align_data/sources/articles/indices.py
mruwnik: maybe something like `contents` would be clearer?

align_data/sources/articles/parsers.py
mruwnik: exactly :D Higher order functions FTW! Or factory classes in this case...

align_data/sources/articles/pdf.py
mruwnik: good catch! I'll update to your function, which should make merging easier

tests/align_data/articles/test_datasets.py
mruwnik: Maybe :/ Articles is starting to grow to be a lot larger than it was initially. I'm not sure what the prettiest structure would be. It makes sense for it to be like this - the only reason the other ones are in folders in for the __init__.py files to contains ready datasets

align_data/sources/articles/html.py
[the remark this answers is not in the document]
Thomas-Lemoine: awesome

align_data/sources/articles/indices.py
Thomas-Lemoine: Ohh my bad I hadn't thought of that difference for the `, '')` vs `) or ''` thing. And hmm, yeah I see the problem. Something to think about

tests/align_data/articles/test_datasets.py
Thomas-Lemoine: Ah I see, yeah that's fair.

align_data/sources/articles/indices.py
mruwnik: yup. Hard won experience has taught me to care about these things. Usually after costing clients large sums of money...

ccstan99 (approved):

PR #144, opened 2023-08-21
Tidy up

mruwnik:
A couple more small bugs:
* handle missing date published in arxiv vanity parser
* ignore values that are None when merging dicts - this was overwiting proper date_published values with Nones

### DIFF
--- a/align_data/sources/articles/pdf.py
+++ b/align_data/sources/articles/pdf.py
@@ -2,6 +2,7 @@ import io
 import logging
 from urllib.parse import urlparse
 from typing import Dict, Any
+from dateutil.parser import ParserError, parse
 
 import requests
 from PyPDF2 import PdfReader
@@ -155,7 +156,10 @@ def parse_vanity(url) -> Dict[str, Any]:
     ]
 
     if date_published := contents.select_one("div.ltx_dates"):
-        date_published = date_published.text.strip("()")
+        try:
+            date_published = parse(date_published.text.strip("()"))
+        except ParserError:
+            "If the date couldn't be parsed, hope that later phases will be more successful"
 
     text = "\n\n".join(
         MarkdownConverter().convert_soup(elem).strip()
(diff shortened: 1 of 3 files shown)


align_data/sources/arxiv_papers.py
```diff
@@ -9,6 +9,13 @@
 logger = logging.getLogger(__name__)
 
 
+def merge_dicts(*dicts):
```
Thomas-Lemoine, line 12: This works but what benefit does it have over the previous implementation?

Thomas-Lemoine (approved):

align_data/sources/arxiv_papers.py
Thomas-Lemoine: Ah I see, thanks

align_data/sources/articles/pdf.py
```diff
@@ -155,7 +156,10 @@ def get_first_child(item):
     ]
 
     if date_published := contents.select_one("div.ltx_dates"):
-        date_published = date_published.text.strip("()")
+        try:
```
Thomas-Lemoine, line 159: just noticed while merging #97 into main that if there's a ParserError, the behaviour 
`"If the date couldn't be parsed, hope that later phases will be more successful"`
is probably a typo. maybe you mean to log it or to return it as an error, but date_published probably ought to be None?

Thomas-Lemoine: maybe just pass and have that as a comment, too

Thomas-Lemoine:
```py
    selected_date = contents.select_one("div.ltx_dates")
    try:
        date_published = parse(selected_date.text.strip("()")) if selected_date else None
    except ParserError:
        # "If the date couldn't be parsed, hope that later phases will be more successful"
        date_published = None
```
?

Thomas-Lemoine: ok I modified it there, #97

PR #146, opened 2023-08-21
Transformer-circuits blog

mruwnik:


ccstan99 (approved):

align_data/sources/blogs/__init__.py
```diff
@@ -56,4 +57,5 @@
         name="deepmind_technical_blog",
         url="https://www.deepmind.com/blog-categories/technical-blogs",
     ),
+    TransformerCircuits(name='transformer-circuits', url='https://transformer-circuits.pub/'),
```
ccstan99, line 60: So this fit with the other blogs but distill format was different?

mruwnik: not really. Distill was left in its own module mainly through inertia. Initially it was quite a lot of unique code, but since most of it could be abstracted away, it probably could be moved over here

Issue #91, opened 2023-07-13
Update readme.md


Thomas-Lemoine:
the readme has some information out of date, like some sources missing, doesn't display our current convention for Keys, etc. We could add more detail in terms of how to get all the sources working, in its own section

Issue #140, opened 2023-08-20
Reorganize source and source_type


ccstan99:
For discussion at next meeting
- Combine individual playlists and channels to a 'youtube' dataset
- Combine separate blogs into 'blogs' dataset
- Combine separate docs into 'special_docs' dataset with source_type as pdf, html, xml, markdown, gdocs, docx, ebook, etc

PR #147, opened 2023-08-21
Bunch up blogs, special_docs and youtube

mruwnik:
https://github.com/StampyAI/alignment-research-dataset/issues/140

The special_docs, pdfs, etc. sources are all bunched together as `special_docs`. Ditto for all blogs as `blogs`, and youtube videos as `youtube`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,63 +4,68 @@ The AI Alignment Research Dataset is a collection of documents related to AI Ali
 
 ## Sources
 
-The following list of sources may change and items may be renamed:
+Here are the list of sources along with sample contents:
 
-- [agentmodels](https://agentmodels.org/)
-- [aiimpacts](https://aiimpacts.org/)
-- [aisafety.camp](https://aisafety.camp/)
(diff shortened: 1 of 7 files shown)


ccstan99 (commented):
Can we store the original/pre-bunched dataset name in another column, i.e. source_type? Plex would like to still allow them to be tracked easily.

mruwnik:
it's already in the meta column, but can be moved to source type. Though that will make things inconsistent, as e.g. arxviv things will have (source, source_type) = ('arxiv', 'pdf'), LW ('lesswrong', 'Greaterwrong') etc., while these would be ('blogs', 'carado.moe') etc., which is pretty much the reverse

ccstan99:
You're right. It probably fine in metadata since we can run sql queries to generate any needed reports.

Thomas-Lemoine (approved):

ccstan99 (approved):
I updated the README file to match what I think the new list of datasets should look like. If it looks ok, I can update HF too. Closes #91

PR #151, opened 2023-08-22
Add AGISF Governance as a data source

mruwnik:
Add a generic airtable datasource, which is used to fetch AGISF data. Unfortunately Airtable doesn't allow anonymous queries, which is annoying. I made a copy of https://airtable.com/appbiNKDcn1sGPGOG/shro9Bx4f2i6QgtTM/tblSicSC1u6Ifddrq as there didn't seem to be a better way of doing it. The downside of that being that it won't be updated automatically. Might be worth reaching out to whoever is maintaing that table for an API key?


README.md
```diff
@@ -7,6 +7,7 @@ The AI Alignment Research Dataset is a collection of documents related to AI Ali
 Here are the list of sources along with sample contents:
 
 - [agentmodel](https://agentmodels.org/)
+- agisf - the recommended readings from the AGI safety foundations
```
ccstan99, line 10: Fix to:
`- [agisf](https://course.aisafetyfundamentals.com/) - recommended readings from AGI Safety Fundamentals`

ccstan99 (approved):

PR #154, opened 2023-08-23
skip entries with falsy field values

Thomas-Lemoine:


align_data/db/models.py
```diff
@@ -125,6 +125,12 @@ def add_meta(self, key, val):
             self.meta = {}
         self.meta[key] = val
 
+    def add_comment(self, comment: str):
```
mruwnik, line 128: perhaps `append_comment`?

align_data/embeddings/pinecone/pinecone_models.py
```diff
@@ -24,6 +32,15 @@ class PineconeEntry(BaseModel):
     text_chunks: List[str]
     embeddings: List[List[float]]
 
+    def __new__(cls, **data):
```
mruwnik, line 35: why `__new__` rather than `__init__`?

This will cause `create_pinecone_vectors(self)` to be all or nothing, as a single article with missing metadata will blow up the whole run. Is that intended?

Thomas-Lemoine: Oh, not at all. Oh, I now realize it's no longer necessary. Previously, I returned None if any field was missing. that meant that if I used __init__, returning None would do nothing (if I understand things right). Using __new__ to return None meant that PineconeEntry(**data) would return None in self._make_pinecone_entry whenever there was a missing field. Now that I deal with all this by handling errors though, I can use __init__, no problem. 

Although, gpt4 gave me the idea for __new__ in the first place, and my understanding was that its purpose was for the dev to be able to not create the object if some condition was detected, since __new__ runs right before __init__? But this guess at how it works seems to be false based on what you're saying, since I don't see how __new__ would affect create_pinecone_vectors in any way, and I certainly don't intend to break the whole run-
Oh, I see what's going on now. I do raise a ValueError, but check in the other file, in _make_pinecone_entry: it tries to create a PineconeEntry, and if there's an error it adds a comment to the article and logs an error and returns None, so the error does not make the whole run fail. Is that what you meant?

mruwnik: Right - my mistake, I didn't notice that. This makes sense, then.

`__init__` is the constructor of an instance, `__new__` more of a factory function. So it sort of goes:

```instance = Class.__new__(params).__init__(params)```

There are cases where this is very useful, same as metaclasses and other dark magic. But in general, the stranger the method, the more you should avoid it :D

**Review by mruwnik — APPROVED** [review]

File: align_data/embeddings/pinecone/pinecone_models.py (hunk 2 above)
Comment (Thomas-Lemoine) [reply align_data/embeddings/pinecone/pinecone_models.py:35]: Makes sense. Previously, using init would have made it break, but now it doesn't, so I will use __init__. However, it appears that pypandic, similarly to dataclasses or wtv, deals with __init__ itself in some way, so I'm a bit worried about modifying __init__ without breaking things

Comment (Thomas-Lemoine) [reply align_data/embeddings/pinecone/pinecone_models.py:35]: looks like it'll become
```py
    def __init__(self, **data):
        """Check for missing (falsy) fields before initializing."""
        missing_fields = [field for field, value in data.items() if not value]
        
        if missing_fields:
            raise MissingFieldsError(f"Missing fields: {missing_fields}")

        super().__init__(**data)
```

Issue #158, opened 2023-08-23
Fix titles


ccstan99:
Strip newlines from titles. Also consider stripping non-alphanumeric chars from the title and "uncase" (all lowercase) when making hash_ids to catch more duplicates.

Thomas-Lemoine:
so there are two ways where changes can be made:
- article.title
- article.generate_id_string()
do you mean that strip new_lines modifies article.title, meaning the displayed article changes, whereas strip non-alphanum and uncase happen in article.generate_id_string(), meaning it won't affect how it's displayed later on?

PR #159, opened 2023-08-24
fix titles

Thomas-Lemoine:
any new article being created will have their article.title key have newlines replaced by spaces

Any new article.generate_id_string() will have its "title" value be stripped of non-alphanumeric values, and lowercased.

#158

### DIFF
--- a/align_data/common/alignment_dataset.py
+++ b/align_data/common/alignment_dataset.py
@@ -1,3 +1,4 @@
+import re
 from datetime import datetime
 from itertools import islice
 import logging
@@ -72,7 +73,7 @@ class AlignmentDataset:
         data = merge_dicts(data, kwargs)
         summary = data.pop("summary", None)
         authors = data.pop("authors", [])
-        data['title'] = (data.get('title') or '').replace('\n', ' ')
+        data['title'] = (data.get('title') or '').replace('\n', ' ').replace('\r', '') or None
 
         article = Article(
             pinecone_update_required=True,
@@ -141,7 +142,7 @@ class AlignmentDataset:
         """Returns a collection of items to be processed."""
         return []
 
-    def get_item_key(self, item):
+    def get_item_key(self, item) -> str:
         """Get the identifier of the given `item` so it can be checked to see whether it's been output.
 
         The default assumption is that the `item` is a Path to a file.
@@ -270,7 +271,7 @@ class MultiDataset(AlignmentDataset):
         for dataset in self.datasets:
             dataset.setup()
 
-    def get_item_key(self, entry):
+    def get_item_key(self, entry) -> str:
         item, dataset = entry
         return dataset.get_item_key(item)
 
--- a/align_data/common/alignment_dataset.py
+++ b/align_data/common/alignment_dataset.py
@@ -72,6 +72,8 @@ class AlignmentDataset:
         data = merge_dicts(data, kwargs)
         summary = data.pop("summary", None)
         authors = data.pop("authors", [])
+        if isinstance((title := data.get("title")), str):
+            data["title"] = title.replace('\n', ' ')
 
         article = Article(
             pinecone_update_required=True,
--- a/align_data/common/alignment_dataset.py
+++ b/align_data/common/alignment_dataset.py
@@ -72,8 +72,7 @@ class AlignmentDataset:
         data = merge_dicts(data, kwargs)
         summary = data.pop("summary", None)
         authors = data.pop("authors", [])
-        if isinstance((title := data.get("title")), str):
-            data["title"] = title.replace('\n', ' ')
+        data['title'] = (data.get('title') or '').replace('\n', ' ')
 
         article = Article(
             pinecone_update_required=True,
--- a/align_data/common/alignment_dataset.py
+++ b/align_data/common/alignment_dataset.py
@@ -149,23 +149,57 @@ class AlignmentDataset:
         """
         return item.name
 
+    @staticmethod
+    def _normalize_url(url: str) -> str:
+        # ending '/'
+        url = url.rstrip("/")
+
+        # Remove http and use https consistently
+        url = url.replace("http://", "https://")
+
+        # Remove www
+        url = url.replace("https://www.", "https://")
+
+        # Remove index.html or index.htm
+        url = re.sub(r'/index\.html?$', '', url)
+
+        # Convert youtu.be links to youtube.com
+        youtube_short_match = re.match(r'https://youtu\.be/([a-zA-Z0-9_-]+)', url)
+        if youtube_short_match:
+            video_id = youtube_short_match.group(1)
+            url = f'https://youtube.com/watch?v={video_id}'
+
+        # Additional rules for mirror domains can be added here
+
+        # agisafetyfundamentals.com -> aisafetyfundamentals.com
+        url = url.replace("https://agisafetyfundamentals.com", "https://aisafetyfundamentals.com")
+
+        return url
+    
+    def _normalize_urls(self, urls: Iterable[str]) -> Set[str]:
+        return {self._normalize_url(url) for url in urls}
+
+
     def _load_outputted_items(self) -> Set[str]:
         """Load the output file (if it exists) in order to know which items have already been output."""
         with make_session() as session:
+            items = set()
             if hasattr(Article, self.done_key):
                 # This doesn't filter by self.name. The good thing about that is that it should handle a lot more
                 # duplicates. The bad thing is that this could potentially return a massive amount of data if there
                 # are lots of items.
-                return set(session.scalars(select(getattr(Article, self.done_key))).all())
+                items =  set(session.scalars(select(getattr(Article, self.done_key))).all())
             # TODO: Properly handle this - it should create a proper SQL JSON select
-            return {item.get(self.done_key) for item in session.scalars(select(Article.meta)).all()}
+            else:
+                items = {item.get(self.done_key) for item in session.scalars(select(Article.meta)).all()}
+            return self._normalize_urls(items)
 
-    def not_processed(self, item):
+    def not_processed(self, item) -> bool:
         # NOTE: `self._outputted_items` reads in all items. Which could potentially be a lot. If this starts to
         # cause problems (e.g. massive RAM usage, big slow downs) then it will have to be switched around, so that
         # this function runs a query to check if the item is in the database rather than first getting all done_keys.
         # If it get's to that level, consider batching it somehow
-        return self.get_item_key(item) not in self._outputted_items
+        return self._normalize_url(self.get_item_key(item)) not in self._outputted_items
 
     def unprocessed_items(self, items=None) -> Iterable:
         """Return a list of all items to be processed.
--- a/align_data/common/alignment_dataset.py
+++ b/align_data/common/alignment_dataset.py
@@ -150,7 +150,10 @@ class AlignmentDataset:
         return item.name
 
     @staticmethod
-    def _normalize_url(url: str) -> str:
+    def _normalize_url(url: str | None) -> str | None:
+        if not url:
+            return url
+
         # ending '/'
         url = url.rstrip("/")
 
--- a/align_data/common/alignment_dataset.py
+++ b/align_data/common/alignment_dataset.py
@@ -167,10 +167,7 @@ class AlignmentDataset:
         url = re.sub(r'/index\.html?$', '', url)
 
         # Convert youtu.be links to youtube.com
-        youtube_short_match = re.match(r'https://youtu\.be/([a-zA-Z0-9_-]+)', url)
-        if youtube_short_match:
-            video_id = youtube_short_match.group(1)
-            url = f'https://youtube.com/watch?v={video_id}'
+        url = url.replace("https://youtu.be/", "https://youtube.com/watch?v=")
 
         # Additional rules for mirror domains can be added here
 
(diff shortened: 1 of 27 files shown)


ccstan99:
This looks good to me but I'd like to have @mruwnik take a look first since we might want to consider how to update current mysql db articles

Thomas-Lemoine:
The test that fails is this:
```py
def test_process_entry():
    dataset = Stampy(name="bla")
    entry = {
        "Question": "Why&NewLine;not&#32;just&#63;",
        "Rich Text": "bla bla bla",
        "UI ID": "1234",
        "Doc Last Edited": "2012-02-03",
    }
    assert dataset.process_entry(entry).to_dict() == {
        "authors": ["Stampy aisafety.info"],
        "date_published": "2012-02-03T00:00:00Z",
        "id": None,
        "source": "bla",
        "source_type": "markdown",
        "summaries": [],
        "text": "bla bla bla",
        "title": "Why\nnot just?",
        "url": "https://aisafety.info?state=1234",
    }
```
It appears that stampy in particular passes the title through a clean_text:
```py
def clean_text(text):
    text = html.unescape(text)
    return re.sub(r"\(/\?state=(\w+)\)", r"(http://aisafety.info?state=\1)", text)
```
and expects the Question's NewLine substrings to be replaced by actual newlines, and for these newlines to stay in the title; this is not the case if we remove newlines from all titles, and so the test_stampy.py test would have to be changed to remove \n from the expected output's 'title'.

Thomas-Lemoine:
this is a brainstorm of the part where you have to modify the `title` and `id` fields of all articles in the db:
```py
from align_data.db.session import make_session

with make_session(auto_commit=True) as session:
    for article in session.query(Article).all():
        if isinstance(article.title, str):
            article.title = article.title.replace("\n", " ").strip()
        article._set_id()
```

align_data/common/alignment_dataset.py
```diff
@@ -72,6 +72,8 @@ def make_data_entry(self, data, **kwargs) -> Article:
         data = merge_dicts(data, kwargs)
         summary = data.pop("summary", None)
         authors = data.pop("authors", [])
+        if isinstance((title := data.get("title")), str):
```
mruwnik, line 75: just do a `data['title'] = (data.get('title') or '').replace('\n', ' ')` - if title is not a string (or None), then something else went very wrong

align_data/db/models.py
```diff
@@ -71,7 +72,16 @@ def __repr__(self) -> str:
         return f"Article(id={self.id!r}, title={self.title!r}, url={self.url!r}, source={self.source!r}, authors={self.authors!r}, date_published={self.date_published!r})"
 
     def generate_id_string(self) -> bytes:
-        return "".join(str(getattr(self, field)) for field in self.__id_fields).encode("utf-8")
```
mruwnik, line 74: you could probably strip from everything. It shouldn't make that much of a difference if it hashes `httpblablabla` rather than `http://bla.bla/bla/`, and that will make the code nicer, e.g.

```
return "".join(
    str(getattr(self, field)).sub(r'[^a-zA-Z0-9\s]', '', val).strip().lower()
    for field in self.__id_fields
).encode("utf-8")
```

or even

```
def normalised(self, field: str) -> str:
   return str(getattr(self, field)).sub(r'[^a-zA-Z0-9\s]', '', val).strip().lower()

def generate_id_string(self) -> bytes:
    return "".join(map(self.normalised, self.__id_fields)).encode("utf-8")
```

mruwnik (commented):
Some tests for this would be good

mruwnik:
> the test_stampy.py test would have to be changed to remove \n from the expected output's 'title'.

Yes, that will have to be changed. But that the point of these tests, so you notice what your changes affect. You could even go so far as to change the tests for the other datasets and add newlines in random places of their titles to make sure they get stripped out

align_data/db/models.py
```diff
@@ -71,7 +72,16 @@ def __repr__(self) -> str:
...
+
+            if field == "title": # remove non-alphanumeric characters, lowercase
+                val = re.sub(r'[^a-zA-Z0-9\s]', '', val).strip().lower()
+            
+            id_str += val
+
+        return id_str.encode("utf-8")
 
     @property
     def __id_fields(self):
```
ccstan99, line 87: @mruwnik What do you think about hashing on url (without title) but allowing for source in summaries docs?
```
    def __id_fields(self):
        if self.source in ["importai", "ml_safety_newsletter", "alignment_newsletter"]:
            return ["url", "source"]
        return ["url"]
 ```

Comment (mruwnik) [reply align_data/db/models.py:87]: should be fine?

**Comment by ccstan99** [comment]
I think we really need to remove the title from the hash_id. We already have tons of duplicates and these should be easy to catch. Oh wait, we might need to strip non-alphanumeric chars from urls too if there's a difference between ending with a '/' and without... and ignore http:// vs https://... and index.html or index.htm. Uggggh.

url | title | source | id | hash_id
-- | -- | -- | -- | --
https://openai.com/charter | [Week 7] OpenAI Charter | agisf | 24216 | 600bc01b97f467b37f16ad8672eaf131
https://openai.com/charter/ | OpenAI Charter | agisf | 24184 | b27183517c304da76d621593b453bf59
https://www.openphilanthropy.org/research/12-tentative-ideas-for-us-ai-policy | “[Week 7] ’12 tentative ideas for US AI policy’ by Luke Muehlhauser” by Luke Muehlhauser | agisf | 24217 | f27f2a05c87891ece9bf6c17449b92f8
https://www.openphilanthropy.org/research/12-tentative-ideas-for-us-ai-policy/ | 12 Tentative Ideas for US AI Policy | agisf | 24172 | 5e0433729c56f89bdd1c0a85b8654167

File: align_data/common/alignment_dataset.py (hunk 1 above)
Comment (Thomas-Lemoine) [reply align_data/common/alignment_dataset.py:75]: sounds good. a few tests will change from this

Comment (Thomas-Lemoine) [reply align_data/common/alignment_dataset.py:75]: and a couple more change from generate_id_string() getting a new behaviour for non-alphanums. I'll make sure I'm only changing ones that are supposed to break 🤷‍♂️ and adding a few

Comment (Thomas-Lemoine) [reply align_data/common/alignment_dataset.py:75]: oh LOL well thanks to tests for catching that that was a typo, "title" is the one getting empty string by default

Comment (Thomas-Lemoine) [reply align_data/common/alignment_dataset.py:75]: the only "title": None remaining is from MediumParser's __call__ which, really, is a parser and so doesn't output an article like the others, and doesn't have its 'title' default set to '', so good sign (and everything passes)

**Comment by Thomas-Lemoine** [comment]
> > the test_stampy.py test would have to be changed to remove \n from the expected output's 'title'.
> 
> Yes, that will have to be changed. But that's the point of these tests, so you notice what your changes affect. You could even go so far as to change the tests for the other datasets and add newlines in random places of their titles to make sure they get stripped out

yeah, that's great :p and yeah I did a little bit of that but I can add some more

File: align_data/db/models.py [hunk 4]
```diff
@@ -72,27 +72,24 @@ def __repr__(self) -> str:
...
+            re.sub(r'[^a-zA-Z0-9\s]', '', str(getattr(self, field))).strip().lower()
+            for field in self.__id_fields
+        ).encode("utf-8")
 
     @property
-    def __id_fields(self):
+    def __id_fields(self) -> List[str]:
+        # if self.source in ["importai", "ml_safety_newsletter", "alignment_newsletter"]:
+        #     return ["url", "source"]
+        # return ["url"]
```
mruwnik: ?

tests/align_data/common/test_alignment_dataset.py
```diff
@@ -163,10 +163,35 @@ def test_data_entry_verify_id_fails():
...
+    entry = dataset.make_data_entry(
+        {
+            "url": "www.arbital.org",
+            "title": "once upon a time",
+            "id": "f2b4e02fc1dd8ae43845e4f930f2d84f",
+        }
+    )
+
+    with pytest.raises(AssertionError):
+        assert entry.generate_id_string() == b"www.arbital.orgonce upon a time"
```
mruwnik, line 193: you don't need the `with pytest.raises` here - the way pytest works is pretty much:

```
def will_test_pass(run_test_func):
    try:
        run_test_func()
        return True
    except AssertionError:
        return False
```
So tests that don't raise an exception pass, those that raise an AssertionError fail, and any other exceptions are bubbled up

align_data/db/models.py
Thomas-Lemoine: https://github.com/StampyAI/alignment-research-dataset/pull/159/commits/1662b81e21b1c51e10e132e0ce4a836d365ffaa3

tests/align_data/common/test_alignment_dataset.py
mruwnik: this test doesn't seem to add much. You could achieve the same by adding an `entry.generate_id_string() != b"www.arbital.orgonce upon a time"` to the previous test, which practically speaking is only testing whether the `==` operation works

tests/align_data/common/test_html_dataset.py
```diff
@@ -155,6 +155,33 @@ def test_html_dataset_process_entry_no_text(html_dataset):
         assert html_dataset.process_entry(article) is None
 
 
+def test_html_dataset_newline_in_title(html_dataset: HTMLDataset):
+    html_with_newline_title = f"""
+    <article>
+        <h1>
+            the title\nwith a newline
```
mruwnik, line 162: you could also just add it, e.g.
```
<h1>
   the title\nwith
  a newline
</h1>
```

align_data/common/alignment_dataset.py
```diff
@@ -72,6 +72,7 @@ def make_data_entry(self, data, **kwargs) -> Article:
         data = merge_dicts(data, kwargs)
         summary = data.pop("summary", None)
         authors = data.pop("authors", [])
+        data['title'] = (data.get('title') or '').replace('\n', ' ')
```
mruwnik, line 75: this might be better as `(data.get('title') or '').replace('\n', ' ').replace('\r', '')`

tests/align_data/common/test_html_dataset.py
Thomas-Lemoine: :D I also vaguely remember hearing that, but no I'm on mac, I'll investigate

align_data/common/alignment_dataset.py
Thomas-Lemoine: alright yeah

Thomas-Lemoine: Do you have a preference for/against adding `or None` at the end of it to make the default `None` instead of `''`? it's only for a bit more testing consistency

mruwnik: don't change code for testing consistency - or at least only if you really, really, really have to.
That being said, I'm all for changing code to be more consistent in general, which this most certainly will be, so go for it! Especially as the field is nullable in the database. Having to deal with both `NULL` and `''` in sql would be annoying

Thomas-Lemoine: that makes sense, and yeah I get a vague sense of horror when I modify the tests so that the tests get passed...

align_data/common/alignment_dataset.py
```diff
@@ -141,30 +142,64 @@ def items_list(self) -> Iterable:
...
+        url = url.replace("http://", "https://")
+
+        # Remove www
+        url = url.replace("https://www.", "https://")
+
+        # Remove index.html or index.htm
+        url = re.sub(r'/index\.html?$', '', url)
+
+        # Convert youtu.be links to youtube.com
+        youtube_short_match = re.match(r'https://youtu\.be/([a-zA-Z0-9_-]+)', url)
```
mruwnik, line 167: wouldn't `url = url.replace('https://youtube.be', 'https://youtube.com/watch?v=')` do the same?

mruwnik (dismissed):

align_data/common/alignment_dataset.py
Thomas-Lemoine: `youtu.be` but yes I think so

Thomas-Lemoine: made the change

align_data/sources/articles/datasets.py
```diff
@@ -116,17 +116,26 @@ def get_contents(self, item) -> Dict:
...
+            if given_url:
+                norm_url = self._normalize_url(given_url)
+                if norm_url in self._outputted_items:
+                    return False
+
+                norm_canonical_url = self._normalize_url(arxiv_canonical_url(given_url))
+                if norm_canonical_url in self._outputted_items:
+                    return False
+
+        return True
```
Thomas-Lemoine, line 139: I could have kept the previous code and added self._normalize_url() around each thing but that seemed messy

Thomas-Lemoine:
ok hopefully this is done with #163 and #158 

after this PR, my understanding is that we rerun all hash_ids, and we recheck for duplicates (which we remove or deal with

after that there should be less duplicates

align_data/sources/articles/datasets.py
mruwnik: you could also try something like 
```
to_check = set((
            self.get_item_key(item), url, source_url , (url and arxiv_cannonical_url(url)), (source_url and arxiv_cannonical_url(source_url))
        ))
return bool(self._outputted_items & to_check)
```
But this is also fine

mruwnik (approved):

mruwnik:
I added a commit to resolve the conflicts

mruwnik (approved):

align_data/sources/articles/datasets.py
Thomas-Lemoine: Oh that's a good one

PR #162, opened 2023-08-25
Fix moderation and add pinecone update with hash ids

henri123lemoine:
- This PR fixes the moderation problem where there could only be 32 chunks moderated per call.
- It also fixes pinecone_update_required stuff, where now it will only be set to True if any of the PineconeMetadata attributes have been modified *and* is_valid is true.

Sorry for again including some cosmetic changes in here, but you guys need to pull from main for your commits to auto-include running black.


mruwnik (dismissed):

align_data/db/models.py
```diff
@@ -120,30 +130,40 @@ def _set_id(self):
...
+                self.title and self.title.strip(),
+                self.authors,
+                self.status == OK_STATUS,
+            ]
         )
 
+        # URL validation
+        try:
+            result = urlparse(self.url)
+            url_check = all([result.scheme in ["http", "https"], result.netloc])
```
Thomas-Lemoine, line 159: result.scheme in ["http", "https"]) and result.netloc`?

align_data/db/session.py
```diff
@@ -32,7 +34,26 @@ def stream_pinecone_updates(
...
         .filter(or_(Article.confidence == None, Article.confidence > MIN_CONFIDENCE))
-        .yield_per(1000)
+    )
+
+
+def get_pinecone_articles_by_ids(
+    session: Session,
+    custom_sources: List[str],
+    force_update: bool = False,
+    hash_ids: List[int] | None = None,
```
Thomas-Lemoine, line 44: how about this requires a list and has no default? like shouldn't it fail if we give it no value as "hash_ids" when running a function that requires them

Thomas-Lemoine (dismissed):

align_data/db/session.py
henri123lemoine: Good idea

henri123lemoine:
The error occurs for unknown reasons. I do not have this error locally when I run pytest.

align_data/db/models.py
```diff
@@ -192,3 +221,4 @@ def to_dict(self):
 
 event.listen(Article, "before_insert", Article.before_write)
 event.listen(Article, "before_update", Article.before_write)
+event.listen(Article, "before_update", Article.check_for_changes)
```
mruwnik, line 224: should this also be `before_insert`?

align_data/db/models.py
```diff
@@ -160,14 +180,23 @@ def before_write(cls, _mapper, _connection, target):
...
-    def to_dict(self):
+    @classmethod
+    def check_for_changes(cls, mapper, connection, target):
+        # Attributes we want to monitor for changes
+        monitored_attributes = list(PineconeMetadata.__annotations__.keys())
+        monitored_attributes.remove("hash_id")
+
+        changed = any(get_history(target, attr).has_changes() for attr in monitored_attributes)
+
+        if changed and target.is_valid:
```
mruwnik, line 196: how about changing this to:
```
if target.is_valid:
   monitored_attributes = ...

  target.pinecone_update_required = changed
```
This will avoid messing about with checking the metadata etc. if not needed

align_data/db/session.py
```diff
@@ -32,7 +34,26 @@ def stream_pinecone_updates(
...
+    """Yield Pinecone entries that require an update and match the given IDs."""
+    if hash_ids is None:
+        hash_ids = []
+
+    yield from (
+        session.query(Article)
+        .filter(or_(Article.pinecone_update_required.is_(True), force_update))
+        .filter(Article.is_valid)
+        .filter(Article.source.in_(custom_sources))
+        .filter(or_(Article.confidence == None, Article.confidence > MIN_CONFIDENCE))
```
mruwnik, line 55: how about a `get_pinecone()` function (preferably with a better name...) with the main query body, e.g.:
```
def get_pinecone(
    session: Session,
    custom_sources: List[str],
    force_update: bool = False,
):
    return (
        session.query(Article)
        .filter(or_(Article.pinecone_update_required.is_(True), force_update))
        .filter(Article.is_valid)
        .filter(Article.source.in_(custom_sources))
        .filter(or_(Article.confidence == None, Article.confidence > MIN_CONFIDENCE))
    )

def get_pinecone_articles_to_update(
    session: Session,
    custom_sources: List[str],
    force_update: bool = False,
):
    yield from get_pinecone(session, custom_source, force_update).yield_per(1000)

def get_pinecone_articles_by_ids(
    session: Session,
    custom_sources: List[str],
    force_update: bool = False,
    hash_ids: List[int] | None = None,
):
     """Yield Pinecone entries that require an update and match the given IDs."""
    yield from get_pinecone(session, custom_source, force_update).filter(Article.id.in_(hash_ids))

File: align_data/settings.py [hunk 6]
```diff
@@ -39,8 +39,19 @@
 host = os.environ.get("ARD_DB_HOST", "127.0.0.1")
 port = os.environ.get("ARD_DB_PORT", "3306")
 db_name = os.environ.get("ARD_DB_NAME", "alignment_research_dataset")
-DB_CONNECTION_URI = f"mysql+mysqlconnector://{user}:{password}@{host}:{port}/{db_name}"
-ARTICLE_MAIN_KEYS = ["id", "source", 'source_type', "title", "authors", "text", "url", "date_published", "status", "comments"]
+DB_CONNECTION_URI = f"mysql+mysqldb://{user}:{password}@{host}:{port}/{db_name}"
```
mruwnik, line 42: change to `f"mysql+mysqlconnector://{user}:{password}@{host}:{port}/{db_name}"` for the tests to start working again

align_data/db/models.py
henri123lemoine: urlparse parses a URL into 6 components: <scheme>://<netloc>/<path>;<params>?<query>#<fragment>
Ensuring that the scheme is one of "http" or "https" is probably unnecessary but might not be, because there are other schemes we wouldn't want to accept.
Regardless, I removed that check for now because I'm not sure that it's right.

align_data/db/models.py
henri123lemoine: I don't think so. It would then miss the times when, e.g., an article's date_published is updated. I guess that shouldn't really occur, but any change to the pinecone metadata should trigger a change to the pinecone database, if we want it to be always up-to-date.

align_data/db/models.py
henri123lemoine: So something like:
```py
@classmethod
def check_for_changes(cls, mapper, connection, target):
    # Attributes we want to monitor for changes
    monitored_attributes = list(PineconeMetadata.__annotations__.keys())
    monitored_attributes.remove("hash_id")

    if target.is_valid:
        changed |= any(get_history(target, attr).has_changes() for attr in monitored_attributes)
        target.pinecone_update_required = changed
```?

Comment (Thomas-Lemoine) [reply align_data/db/models.py:196]: why not just:
```py
@classmethod
def check_for_changes(cls, mapper, connection, target):
    if not target.is_valid:
        return
    monitored_attributes = list(PineconeMetadata.__annotations__.keys())
    monitored_attributes.remove("hash_id")
    changed |= any(get_history(target, attr).has_changes() for attr in monitored_attributes)
    target.pinecone_update_required = changed
```
?

Thomas-Lemoine (approved):

Issue #160, opened 2023-08-24
Deduplicate alignmentforum & lesswrong


ccstan99:
Check if a title & author already exists in alignmentforum, don't add to lesswrong. Remove any existing duplicates. 
Here's a [list items from both datasets sorted by titles ](https://docs.google.com/spreadsheets/d/1aZcDOva2JW6_0TtQiVi-1Jlxn3dOv-PMNxvRJ7_cxTY/edit#gid=102716371)to see duplicated samples.

Issue #156, opened 2023-08-23
Automatic indeces marking


markovial:
Mark sources from indeces or newsletter suggestions to make sure that the places that the sources came from are tracked in the database.

PR #167, opened 2023-08-25
Tidy up

mruwnik:
* change the name of the DB docker container to stampy-db
* [add the initial source for indices articles](https://github.com/StampyAI/alignment-research-dataset/issues/156)
*[ skip greaterwrong cross posts](https://github.com/StampyAI/alignment-research-dataset/issues/160)

### DIFF
--- a/align_data/settings.py
+++ b/align_data/settings.py
@@ -39,7 +39,7 @@ password = os.environ.get("ARD_DB_PASSWORD", "we all live in a yellow submarine"
 host = os.environ.get("ARD_DB_HOST", "127.0.0.1")
 port = os.environ.get("ARD_DB_PORT", "3306")
 db_name = os.environ.get("ARD_DB_NAME", "alignment_research_dataset")
-DB_CONNECTION_URI = f"mysql+mysqldb://{user}:{password}@{host}:{port}/{db_name}"
+DB_CONNECTION_URI = f"mysql+mysqlconnector://{user}:{password}@{host}:{port}/{db_name}"
 ARTICLE_MAIN_KEYS = ["id", "source", 'source_type', "title", "authors", "text", "url", "date_published", "status", "comments"]
 
 ### EMBEDDINGS ###
--- a/align_data/sources/articles/indices.py
+++ b/align_data/sources/articles/indices.py
@@ -60,7 +60,7 @@ def aisafetysupport():
 
 def format_mlsafety_course(a):
     if (a.get("href") or "").startswith("http"):
-        return {"title": a.text, "url": a.get("href")}
+        return {"title": a.text, "url": a.get("href"), "initial_source": "mlsafety_course"}
 
 
 def format_anthropic(post):
@@ -75,6 +75,7 @@ def format_anthropic(post):
         "title": get_text(post, "div.post-heading"),
         "url": url,
         "source_url": source_url,
+        "initial_source": "anthropic",
         "date_published": date_published,
     }
 
@@ -84,6 +85,7 @@ def format_safe_ai(item):
         "title": get_text(item, "h4"),
         "url": item.find("a").get("href"),
         "source_url": item.find("a").get("href"),
+        "initial_source": "safe.ai",
         "authors": get_text(item, "h4 ~ p"),
     }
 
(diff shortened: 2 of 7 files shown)


align_data/settings.py
```diff
@@ -39,7 +39,7 @@
 host = os.environ.get("ARD_DB_HOST", "127.0.0.1")
 port = os.environ.get("ARD_DB_PORT", "3306")
 db_name = os.environ.get("ARD_DB_NAME", "alignment_research_dataset")
-DB_CONNECTION_URI = f"mysql+mysqldb://{user}:{password}@{host}:{port}/{db_name}"
+DB_CONNECTION_URI = f"mysql+mysqlconnector://{user}:{password}@{host}:{port}/{db_name}"
```
mruwnik, line 42: this is the oracle driver. Might be better? At least it doesn't need external libraries to be installed

align_data/sources/greaterwrong/greaterwrong.py
```diff
@@ -79,8 +84,29 @@ def setup(self):
...
+        with make_session() as session:
+            articles = (
+                session
+                .query(Article.url, Article.title, Article.authors)
+                .where(Article.source_type == self.source_type)
+                .all()
+            )
+            return (
+                {a.url for a in articles},
+                {(a.title.replace('\n', '').strip(), a.authors) for a in articles},
```
mruwnik, line 98: fetch `(url, (title, authors))` pairs to check against new items to be processed

local_db.sh
```diff
@@ -1,10 +1,10 @@
 #!/usr/bin/env bash
 ROOT_PASSWORD=my-secret-pw
 
-docker start alignment-research-dataset
+docker start stampy-db
```
mruwnik, line 4: this might require `docker container rename alignment-research-dataset stampy-db` to keep the local database. The idea being that there will be a single container for all stampy db stuff (ARD + chatbot), rather than seperate ones

align_data/sources/greaterwrong/greaterwrong.py
Thomas-Lemoine: Am I right in understanding that this makes it so pages wont be fetched if, in the db, there is another greaterwrong source that has both the same authors and same title, to avoid sub-greaterwrong duplicates? 
My followup question is whether authorship transfers that neatly between the accounts, or it requires all authors to have chosen the same exact name twice. but this is nice

local_db.sh
Thomas-Lemoine: oh that's great

Thomas-Lemoine (approved):

align_data/sources/greaterwrong/greaterwrong.py
mruwnik: correct. People mostly seem to have the same name, so while this won't be perfect, it should greatly reduce the amount of duplicates

align_data/sources/articles/indices.py
```diff
@@ -94,6 +96,7 @@ def format_far_ai(item):
         "url": f'https://www.safe.ai/research{item.select_one(".article-title a").get("href")}',
```
ccstan99, line 96: Why is this url safe.ai in the format_far_ai function? They're separate entities.

align_data/sources/articles/indices.py
```diff
@@ -75,6 +75,7 @@ def format_anthropic(post):
         "title": get_text(post, "div.post-heading"),
         "url": url,
         "source_url": source_url,
+        "initial_source": "anthropic",
```
ccstan99, line 78: We can't pass the url to the formatter to use for "initial_source" instead of hardcoding all of these?
```
def indice_fetcher(url, main_selector, item_selector, formatter):
```

mruwnik: you mean e.g. `"https://www.anthropic.com/research"`?

ccstan99: Not necessarily that specific example just in general "initial_source": <whatever the url was passed to indice_fetcher instead of hardcoding for each type>. Isn't there a simpler way to do that? Not a priority, just thought it'd be easier for maintenance.

mruwnik: it's already sort of hard coded. Each indice has a dedicated function for it, with its url hardcoded. In theory, if there were multiple websites using the same code, it would be more generic, but it practice it's hardcoded. That being said, using the main url is a good idea

PR #174, opened 2023-08-31
Arbital refactor

Thomas-Lemoine:
modified a few functions, especially markdownify_text, so that summaries are saved as summaries rather than as part of the text
+ tests

### DIFF
--- a/align_data/sources/arbital/arbital.py
+++ b/align_data/sources/arbital/arbital.py
@@ -66,7 +66,7 @@ def markdownify_text(current: List[str], view: Iterator[Tuple[str, str]]) -> Tup
         current = []
         view = iter([('[', 'summary: A behaviorist '), ('summary: A behaviorist ', '['), ('[', '6w genie'), ('6w genie', ']'), (']', ']'), (']', None)])
     The function should return:
-        ('summary: A behaviorist [genie](https://arbital.com/p/6w)', '')
+        ('A behaviorist [genie](https://arbital.com/p/6w)', '')
     
     Note:
     This function assumes that `view` provides a valid Arbital markdown sequence. Malformed sequences might lead to 
@@ -78,8 +78,8 @@ def markdownify_text(current: List[str], view: Iterator[Tuple[str, str]]) -> Tup
     for part, next_part in view:
         if part == "[":
             # Recursively try to parse this new section - it's probably a link, but can be something else
-            sum, text = markdownify_text([part], view)
-            summary += sum + "\n\n"
+            sub_summary, text = markdownify_text([part], view)
+            summary += sub_summary + "\n\n"
             current.append(text)
 
         elif part == "]":
@@ -88,9 +88,11 @@ def markdownify_text(current: List[str], view: Iterator[Tuple[str, str]]) -> Tup
                 current.append(part)
                 in_link = True
             else:
-                # this is the arbital summary - just join it for now, but it'll have to be handled later
+                # this is the arbital summary - we return it as a summary, 
+                # and strip characters from the summary's text
                 if current[1].startswith("summary"):
-                    return "".join(current[1:]), ""
+                    summary_tag, summary = "".join(current[1:]).split(":", 1)
+                    return summary_tag + ": " + summary.strip(), ""
                 # if this was a TODO section, then ignore it
                 if current[1].startswith("todo"):
                     return "", ""
--- a/align_data/sources/arbital/arbital.py
+++ b/align_data/sources/arbital/arbital.py
@@ -1,5 +1,9 @@
 import re
+from dataclasses import dataclass, field
+from datetime import datetime, timezone
 import logging
+from typing import List, Tuple, Iterator, Dict, Union, Any, TypedDict
+
 import requests
 from datetime import datetime, timezone
 from dateutil.parser import parse
@@ -10,19 +14,29 @@ from dataclasses import dataclass
 logger = logging.getLogger(__name__)
 
 
-def parse_arbital_link(contents):
-    text = contents[1].split(" ")
-    url = f"https://arbital.com/p/{text[0]}"
-    if len(text) > 1:
-        title = " ".join(text[1:])
-    else:
-        title = url
+class Page(TypedDict, total=False):
+    likeableId: str
+    likeableType: str
+    title: str
+    editCreatedAt: str
+    pageCreatedAt: str
+    alias: str
+    userId: str
+    tagIds: str
+    changeLogs: List[Dict[str, Any]] 
+
+
+def parse_arbital_link(contents: List[str]) -> str:
+    page_id, *title_parts = contents[1].split(" ")
+    url = f"https://arbital.com/p/{page_id}"
+    title = " ".join(title_parts) if title_parts else url
     return f"[{title}]({url})"
 
 
-def flatten(val):
+def flatten(val: Union[List[str], Tuple[str], str]) -> List[str]:
+    """Flattens a nested list."""
     if isinstance(val, (list, tuple)):
-        return [item for i in val for item in flatten(i)]
+        return [item for sublist in val for item in flatten(sublist)]
     return [val]
 
 
@@ -84,7 +98,7 @@ def markdownify_text(current, view):
     return summary.strip(), "".join(flatten(current)).strip()
 
 
-def extract_text(text):
+def extract_text(text: str) -> Tuple[str, str]:
     parts = [i for i in re.split(r"([\[\]()])", text) if i]
     return markdownify_text([], zip(parts, parts[1:] + [None]))
 
@@ -106,10 +120,10 @@ class Arbital(AlignmentDataset):
         "sec-fetch-dest": "empty",
         "accept-language": "en-US,en;q=0.9",
     }
-    titles_map = {}
+    titles_map: Dict[str, str] = field(default_factory=dict)
 
     @property
-    def items_list(self):
+    def items_list(self) -> List[str]:
         logger.info("Getting page aliases")
         items = [
             alias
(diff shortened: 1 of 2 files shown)


Thomas-Lemoine:
I fetched a bunch of arbital articles to check that it worked correctly and it seems fine, but occasionally there is stuff like:
```
The reasoning for an instrumental convergence claim says that for many utility functions $U_k$ and situations $S_i$ a $U_k$-consequentialist in situation $S_i$ will probably find some best policy $\pi_k = \underset{\pi_i \in \Pi}{\operatorname{argmax}}  \mathbb E [U_k | S_i, \pi_i ](https://arbital.com/p/)$ that happens to be inside the partition $X$.  If instead in situation $S_k$...
```
Where some (https://arbital.com/p/) comes out of nowhere

Thomas-Lemoine:
my guess is that this would also be a problem for the current dataset though, since I don't see how that behaviour would have changed with my new code. basically formulas that use brackets will confuse the parser

align_data/sources/arbital/arbital.py
```diff
@@ -10,81 +14,111 @@
...
+def markdownify_text(current: List[str], view: Iterator[Tuple[str, str]]) -> Tuple[str, str]:
+    """
+    Recursively parse text segments from `view` to generate a markdown Abstract Syntax Tree (AST).
+    
+    This function helps in transitioning from Arbital's specific markdown extensions to standard markdown. It specifically
+    handles two main features:
+    - "[summary: <contents>]" blocks, which are used in Arbital to add summaries.
+    - "[123 <title>]" which are Arbital's internal links pointing to https://arbital.com/p/123, with link title <title>.
+    
+    Args:
```
mruwnik, line 52: I see you're doing proper docstring contents - good :D

Is there a reason why you've formatted the arguments like this? The previous way (`:param <type> <name>: <desc>`) is the [reStructured format](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#python-signatures) which is the official (I think? Maybe?) format for specifying paramater documentation. I'm not saying that you should stick to it, just that you should know why you're not

mruwnik:
That's [mathjax](https://www.mathjax.org/) code. It should be fine. Might even be worth adding mathjax to the chatbot and seeing if it can generate pretty equations?

align_data/sources/arbital/arbital.py
```diff
@@ -10,81 +14,111 @@
...
+    title = " ".join(title_parts) if title_parts else url
     return f"[{title}]({url})"
 
 
-def flatten(val):
+def flatten(val: Union[List[str], Tuple[str], str]) -> List[str]:
+    """Flattens a nested list."""
     if isinstance(val, (list, tuple)):
-        return [item for i in val for item in flatten(i)]
+        return [item for sublist in val for item in flatten(sublist)]
```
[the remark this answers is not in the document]
henri123lemoine: forget it

align_data/sources/arbital/arbital.py
```diff
@@ -10,81 +14,111 @@
...
+            if next_part == "(":
+                # mark that it's now in the url part of a markdown link
+                current.append(part)
+                in_link = True
+            else:
+                # this is the arbital summary - we return it as a summary, 
+                # and strip characters from the summary's text
+                if current[1].startswith("summary"):
+                    summary_tag, summary = "".join(current[1:]).split(":", 1)
+                    return summary_tag + ": " + summary.strip(), ""
```
mruwnik, line 95: this will cause all summary entries to be prefixed with `summary:` which seems a bit pointless

Thomas-Lemoine:
> That's [mathjax](https://www.mathjax.org/) code. It should be fine. Might even be worth adding mathjax to the chatbot and seeing if it can generate pretty equations?

You might be misunderstanding what I meant when I quoted that result. somewhere in that answer, which corresponds to this I think: https://arbital.com/p/instrumental_convergence/
there's a usage of mathjax code, but our parsers sees it as brackets and treats it like a link ([123 <title>], but since the content of the brackets is not in that desired format, the parser messes up. It tries to create [title](link), I think? I'm confused.

Thomas-Lemoine:
One thing I'm considering is that if `parse_arbital_link` tries to create a link taht is empty (ie `(https://arbital.com/p/)`), we just ignore that, close the bracket as though it were a real bracket, and go on our merry way. 

Except, it seems like there are two cases that idk how we could distinguish: Either there's math stuff in brackets, in which case we want to keep the brackets as is and show the contents of the brackets, OR there's non-finished links, ie stuff like [pseudoconsequentialist], where someone links to a page not yet created in the hopes that it gets created in the future and can be automatically replaced by a real link. In those cases, we probably want to remove the brackets. Does that make sense?

align_data/sources/arbital/arbital.py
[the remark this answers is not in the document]
Thomas-Lemoine: there's also 
```py
from itertools import chain
flattened_list = chain.from_iterable(lst)
```
I believe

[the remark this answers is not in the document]
Thomas-Lemoine: nvm doesn't work. This is good enough and simple enough imo

align_data/sources/arbital/arbital.py
Thomas-Lemoine: The reason is that gpt4 wrote that part. changing it back rn

Thomas-Lemoine: done

align_data/sources/arbital/arbital.py
Thomas-Lemoine: Yeah, I wasn't sure about that. The point of it was this: sometimes, in the same post there are many summaries, some more technical and some less technical, I think? Like this:
`[summary(Technical): The phrase "disjoint union" is used to mean one of two slightly different operations on [3jz sets]. It indicates either "just take the [5s8 union], but also notice that we've been careful to ensure that the sets have empty [5sb intersection]", or "perform a specific operation on each set to ensure that the resulting sets have empty intersection, and then take the union".]`
became:
```
summary(Technical): The phrase "disjoint union" is used to mean one of two slightly different operations on [sets](https://arbital.com/p/3jz). It indicates either "just take the [union](https://arbital.com/p/5s8), but also notice that we've been careful to ensure that the sets have empty [intersection](https://arbital.com/p/5sb)", or "perform a specific operation on each set to ensure that the resulting sets have empty intersection, and then take the union".
```

Otherwise, there's just a bunch of summaries one after the other and no context as to what differentiates them. That's the overall idea for why I kept the summary: thing, because it also doubles as
summary(Technical): or summary(whatever):, but there might be a better alternative, or I can remove the summary_tag altogether

Thomas-Lemoine: One possibility is to keep the tag (a summary: doesn't harm us much, but it has the added benefit of being found more easily if the user asks for a summary. although, that should be a pretty rare occurrence), but have each separate summary be split into lists, and then return summary as a list of strings

Thomas-Lemoine: nvm, summary is a str

Thomas-Lemoine:
`For some more flavorful examples of this method of using Bayes' rule, see [https://www.gwern.net/docs/statistics/1994-falk The ups and downs of the hope function in a fruitless search].`
becomes
`For some more flavorful examples of this method of using Bayes' rule, see [The ups and downs of the hope function in a fruitless search](https://arbital.com/p/https://www.gwern.net/docs/statistics/1994-falk).`

Thomas-Lemoine:
handled an additional edge case for `parse_arbital_link`

align_data/sources/arbital/arbital.py
Thomas-Lemoine: If you have a preference for list of summaries not having any summary tag they are easy to remove. It will appear like a listing of different paragraphs that summarize the article slightly differently. that would be good enough

mruwnik: summary is a single string, but check `make_data_entry` - there it takes the summary and appends it to the resulting article's `summaries` list. You can work with that. Either by passing in `summaries` rather than `summary` and adding extra code to handle it in the base `make_data_entry`, or by doing something in the arbital class

mruwnik (approved):
I'd prefer multiple summary rows being added, but this will also work

Thomas-Lemoine:
I'll make another PR that changes entries given to make_data_entry; I guess it'll be a `summaries`: List[str] then? it feels like it used to be that though, so do you remember why you switched it back? I suppose each entry being given a "summary" seems a bit more intuitive than giving it a list of many summaries, but also if we allow for an article to have many summaries, we might as well let them all be created from the same entry.

mruwnik:
Most things that have summaries (e.g. arxiv or the alignment newsletter) only have a single summary, so it was easier to do it that way. Which of course breaks here :D

Thomas-Lemoine:
That makes sense. I suppose I can just replace "summary" key with "summaries" and the string with [string], but yeah that makes it a bit less obvious to use. alternatively, the "summary" key might take either List[str] | str, and then it checks if it's an instance of str or of list and decides accordingly, and that's maybe most flexible but maybe also more confusing or error prone? Not sure. Or, two keys, one "summary" and one "summaries", the summary one has str and summaries has List[str]; we assume only one is given; and in `make_data_entry`, the "summary" string is turned into a List[str], `[string]`, appended to "summaries" or wtv, and then it's more flexible but has two different keys depending on how you want to pass in the summary, either as a list of strings or a singular string

mruwnik:
I'd go with checking for both keys, and even going so far as to join them if both are provided

Thomas-Lemoine:
yeah, makes sense. Will add that shortly

PR #176, opened 2023-09-01
fix 2048+ batch embedding

henri123lemoine:
Should solve the errors caused by `assert len(texts) <= 2048`.

### DIFF
--- a/align_data/embeddings/embedding_utils.py
+++ b/align_data/embeddings/embedding_utils.py
@@ -88,34 +88,39 @@ def handle_openai_errors(func):
 
 
 @handle_openai_errors
+def _single_batch_moderation_check(batch: List[str]) -> List[ModerationInfoType]:
+    """Process a batch for moderation checks."""
+    return openai.Moderation.create(input=batch)["results"]
+
+
 def moderation_check(texts: List[str], max_texts_num: int = 32) -> List[ModerationInfoType]:
-    """
-    Check moderation on a list of texts.
-
-    Parameters:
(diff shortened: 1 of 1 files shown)


align_data/embeddings/embedding_utils.py
```diff
@@ -88,34 +88,39 @@ def wrapper(*args, **kwargs):
...
-    - List[ModerationInfoType]: List of moderation results for the provided texts.
-    """
-    total_texts = len(texts)
-    results = []
+def moderation_check(texts: List[str], max_texts_num: int = 32) -> List[ModerationInfoType]:
+    """Batch moderation checks on list of texts."""
+    return [
+        result
+        for batch in (texts[i : i + max_texts_num] for i in range(0, len(texts), max_texts_num))
+        for result in _single_batch_moderation_check(batch)
```
mruwnik, line 101: this could be abstracted away into a generic process-in-batches function, i.e.

```
def process_in_batches(items, batch_size, process_fn):
    return [
       result
       for batch in (items[i : i + batch_size] for i in range(0, len(items), batch_size))
      for result in process_fn(batch)
  ]
```
But the current way is also totally fine

mruwnik (approved):

align_data/embeddings/embedding_utils.py
henri123lemoine: yeah, that's a good point

henri123lemoine: I can make the change at some point when there are no other priorities

PR #178, opened 2023-09-01
Adding back summaries entry key

Thomas-Lemoine:
This makes it so process_entry can pass in, to make_data_entry, either a "summary" key with type string, a "summaries" key with type List[str], or both.
Adds flexibility and lets us give a single summary for most sources, but a list of summaries for arbital (to be implemented)

### DIFF
--- a/align_data/common/alignment_dataset.py
+++ b/align_data/common/alignment_dataset.py
@@ -71,7 +71,11 @@ class AlignmentDataset:
 
     def make_data_entry(self, data, **kwargs) -> Article:
         data = merge_dicts(data, kwargs)
+
+        summaries = data.pop("summaries", [])
         summary = data.pop("summary", None)
+        summaries += [summary] if summary else []
+
         authors = data.pop("authors", [])
         data['title'] = (data.get('title') or '').replace('\n', ' ').replace('\r', '') or None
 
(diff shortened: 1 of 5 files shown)


Thomas-Lemoine:
Ah, I had not envisioned that the "files changed" portion would not be synched with the fact that these changes were added to main. I just created a branch that based itself on arbital-changes, thinking that, since arbital_changes was now merged into main, doing so would be equivalent to branching off a new, fetched, main branch. Understanding why those things are not equivalent is testing my understanding of git a bit

Thomas-Lemoine:
ok yeah there were changes to main that weren't part of my branch, which makes more sense. now this branch's change is more easily visible

align_data/common/alignment_dataset.py
```diff
@@ -80,7 +84,7 @@ def make_data_entry(self, data, **kwargs) -> Article:
             **{k: v for k, v in data.items() if k in ARTICLE_MAIN_KEYS},
         )
         self._add_authors(article, authors)
-        if summary:
+        for summary in summaries: # Note: This will be skipped if summaries is empty
             article.summaries.append(Summary(text=summary, source=self.name))
```
mruwnik, line 88: [NIT] I'd personally go with `article.summaries += [Summary(text=summary, source=self.name) for summary in summaries)]`

mruwnik (approved):

mruwnik (dismissed):

mruwnik (approved):

PR #179, opened 2023-09-01
Arbital many summaries

Thomas-Lemoine:


align_data/common/alignment_dataset.py
```diff
@@ -71,7 +71,11 @@ def _add_authors(self, article: Article, authors: List[str]) -> Article:
 
     def make_data_entry(self, data, **kwargs) -> Article:
         data = merge_dicts(data, kwargs)
+
```
Thomas-Lemoine, line 74: note: these changes are from the other PR, so I think they can be ignored?

align_data/common/alignment_dataset.py
```diff
@@ -71,7 +71,11 @@ def _add_authors(self, article: Article, authors: List[str]) -> Article:
...
+        summaries = data.pop("summaries", [])
         summary = data.pop("summary", None)
+        summaries += [summary] if summary else []
```
Thomas-Lemoine, line 77: these changes are from the other PR, so iiuc they can be ignored

tests/align_data/sources/test_arbital.py
```diff
@@ -86,37 +86,45 @@ def test_markdownify_text_contents_arbital_markdown(text, expected):
...
-            ("summary: summaries should be extracted", "bla bla bla"),
+            (["summary: summaries should be extracted"], "bla bla bla"),
+        ),
+        (
+            "[summary: summaries should be extracted] [summary(Technical): technical summary should be handled separately] bla bla bla",
+            (["summary: summaries should be extracted", "summary(Technical): technical summary should be handled separately"], "bla bla bla"),
+        ),
+        (
+            "[summary: summaries should be extracted] bla bla bla [summary(Technical): summaries should work in the middle too] bla bla bla",
+            (["summary: summaries should be extracted", "summary(Technical): summaries should work in the middle too"], "bla bla bla  bla bla bla"),
```
Thomas-Lemoine, line 97: I'm not sure about this behaviour. If a summary is in the middle of the text, it already works (did not need to modify the code for this). However, this is an accidental thing, but the text before and after the summary brackets appears to be separated by an additional space? So
`"[summary: bla] bla [summary: bla] bla"` will result in the text being `"bla  bla"` (notice the two spaces). Seems fine but a bit arbitrary

Thomas-Lemoine: I suppose it's due to there being a space between the text and the following summary, and between the summary and the following text? Will verify

Thomas-Lemoine: yep, seems like that's right. when there's no space between the text and the opening bracket of the summary, that space between the two texts disappears

Thomas-Lemoine: It's kind of odd when any summary is after a text. Maybe we want to indicate the separation between the two texts, or some other thing. But this would happen very rarely or never

mruwnik: should be fine. If it causes problems, it can be fixed then. Though you could add a comment to mark that - like a TODO or FIXME

mruwnik (approved):

PR #180, opened 2023-09-03
Validate MIN_CONFIDENCE

mruwnik:
Properly handle MIN_CONFIDENCE and make sure it has a correct value.

One downside of this approach is that MIN_CONFIDENCE isn't needed when parsing stuff, only when adding to pinecone. So MIN_CONFIDENCE will be validated even when not needed. The alternative would be to have a getter function (or something), or to check it in each place it's used. The first option adds another layer of abstraction, while the second one will at some point be forgotten and there will be a flow that uses it but doesn't check it (Moore's law)

### DIFF
--- a/align_data/db/session.py
+++ b/align_data/db/session.py
@@ -44,7 +44,7 @@ def get_pinecone_articles_by_sources(
 
 def get_pinecone_articles_by_ids(
     session: Session,
-    hash_ids: List[int],
+    hash_ids: List[str],
     force_update: bool = False,
 ):
     return get_pinecone_articles(session, force_update).filter(Article.id.in_(hash_ids))
--- a/.env.example
+++ b/.env.example
@@ -1,12 +1,17 @@
 LOG_LEVEL="INFO"
+MIN_CONFIDENCE="0.5"
+
 CODA_TOKEN=""
+YOUTUBE_API_KEY=""
+
 ARD_DB_USER="user"
 ARD_DB_PASSWORD="we all live in a yellow submarine"
 ARD_DB_HOST="127.0.0.1"
 ARD_DB_PORT="3306"
 ARD_DB_NAME="alignment_research_dataset"
+
 OPENAI_API_KEY="sk-XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
+
 PINECONE_INDEX_NAME="stampy-chat-ard"
 PINECONE_API_KEY="xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
 PINECONE_ENVIRONMENT="xx-xxxxx-gcp"
-YOUTUBE_API_KEY=""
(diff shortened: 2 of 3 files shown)


align_data/db/session.py
```diff
@@ -44,7 +44,7 @@ def get_pinecone_articles_by_sources(
 
 def get_pinecone_articles_by_ids(
     session: Session,
-    hash_ids: List[int],
+    hash_ids: List[str],
```
Thomas-Lemoine, line 47: nice catch

Thomas-Lemoine:
As I understand it, this makes it so, if an article is set to a confidence lower than, say, 0.5 (if MIN_CONFIDENCE is set to 0.5), the article will be excluded from pinecone. In my mind, MIN_CONFIDENCE would be 0, so that stuff with 0 confidence is excluded, and for the rest, their "quality" as a source is on the range from 0 to 1, where stuff imperceptibly close to 0 would be found in the top-k options in pinecone excessively rarely, and stuff imperceptibly close to 1 would be very encouraged there (though I don't know exactly how we'll implement that, so 1 might be 5x boosted or 100x boosted in the search, I don't know)

mruwnik:
Why not both? :D

Thomas-Lemoine:
I'm not sure what you mean by both

mruwnik:
there can be a hard cutoff for which data is not added to pinecone (e.g. the EY story about the [isekaied girl](https://eliezeryudkowsky.gumroad.com/l/GirlCorrupted)), but the confidence can also be used as a weight

Thomas-Lemoine:
Oh, so you might think a source contains good info for a chatbot answer with 10% probability, and still want it not to be embedded because you'd only want to keep the ones you think are going to be useful with probability 50% and above?

Thomas-Lemoine:
I suppose I'm confused as to what the default is, if confidence is not added? presumably, anything below that default, but above MIN_CONFIDENCE, would still be in the search but would be "penalized" compared to the average source, whereas anything between CONFIDENCE of default and CONFIDENCE of 1 would be boosted. 

I can run you by an idea: Instead of confidence, we have a more evocative "search-boost", or "boost_factor", or something along those lines, which is a float (-inf, +inf) which indicates how much more likely than usual it would be to find this source during a standard search. Meaning: for sources that previously had confidence = 0, the boost_factor would be -inf, which hopefully is a fine float to store. For some articles that seem to have weak content, the boost factor could be negative, like -5, meaning it is 5x less likely to be a result of the semantic search; and if an article is particularly good, like a stampy.info Q&A answer, the boost_factor could be 10x, which may be more intuitive than a confidence of 1, which might boost by some unclear quantity the likelihood that stampy.info answers are used.

Maybe we could repurpose confidence for something like this, or just associate some confidence value with some boost_factor (ie, if an article has a CONFIDENCE of 0.99, it is implicitely boosted by 5x, or something, but if an article has a confidence under 0, or under -3, it is just not embedded).

Thomas-Lemoine:
An issue of course is how to create a boost like this, and last time I checked, we paused using any kind of bias on the length of the embedding vectors since that didn't seem to work too well in our testing.

mruwnik:
`confidence = None` could be mapped to `confidence = MIN_CONFIDENCE`, or even just have a default confidence that is used. But yes, this is confusing and should probably be properly discussed

Thomas-Lemoine (approved):
I'll accept this for now in any case

PR #181, opened 2023-09-03
Update pinecone

mruwnik:
* Change `pinecone_update_required` to be an enum, so it can store more states
* While updating pinecone embeddings, remove any articles from pinecone that are invalid or marked for deletion

Practically speaking, this means that running the update Pinecone command will both add embeddings for missing articles, but also remove embeddings for any articles that are invalid


align_data/db/session.py
```diff
@@ -25,29 +25,43 @@ def make_session(auto_commit=False):
...
-        .filter(or_(Article.confidence == None, Article.confidence > MIN_CONFIDENCE))
+        .filter(or_(Article.confidence == None, Article.confidence >= MIN_CONFIDENCE))
+    )
+
+
+def get_pinecone_articles_to_remove(session: Session):
+    return (
+        session.query(Article)
+        .filter(or_(
+            Article.pinecone_status == PineconeStatus.pending_removal,
```
mruwnik, line 42: no `force_update` for this, as that would result in pinecone being cleared out

align_data/db/session.py
```diff
@@ -25,29 +25,43 @@ def make_session(auto_commit=False):
...
         session.query(Article)
-        .filter(or_(Article.pinecone_update_required.is_(True), force_update))
+        .filter(or_(Article.pinecone_status.in_(statuses), force_update))
         .filter(Article.is_valid)
-        .filter(or_(Article.confidence == None, Article.confidence > MIN_CONFIDENCE))
+        .filter(or_(Article.confidence == None, Article.confidence >= MIN_CONFIDENCE))
+    )
+
+
+def get_pinecone_articles_to_remove(session: Session):
```
mruwnik, line 38: this will return any and all articles that shouldn't be in pinecone, i.e. marked for deletion, invalid or with low confidence

align_data/db/models.py
```diff
@@ -66,7 +75,7 @@ class Article(Base):
     status: Mapped[Optional[str]] = mapped_column(String(256))
     comments: Mapped[Optional[str]] = mapped_column(LONGTEXT)  # Editor comments. Can be anything
 
-    pinecone_update_required: Mapped[bool] = mapped_column(Boolean, default=False)
+    pinecone_status: Mapped[PineconeStatus] = mapped_column(Enum(PineconeStatus), default=PineconeStatus.absent)
```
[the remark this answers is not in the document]
Thomas-Lemoine: Is the following right:
That way we distinguish articles that 
1) have just been fetched, and should be added to pinecone
2) have just been fetched, and shouldn't be added to pinecone because they're not valid
3) been fetched a while ago, but shouldn't be added to pinecone because they're already in pinecone
so that we can distinguish the case 2 from 3, so mysql stores the ground truth of if an article is actually in the pinecone db or not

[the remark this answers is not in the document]
mruwnik: exactly. Plus the additional option of manually specifying that a given article should be removed from pinecone even though it's valid and has high confidence (dunno why, but it's possible). 
So:
1. pending_addition - has just been fetched, is valid, so should be added soon
2. absent - isn't in pinecone and shouldn't be added (either was invalid, or was manually removed) - this is the default state
3. added - is valid and has been added to pinecone
4. pending_removal - was in pinecone, but someone decided it shouldn't be (for whatever reason) and so will be removed as soon as the pinecone updater gets round to it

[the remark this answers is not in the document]
mruwnik: MySQL doesn't so much store the ground truth of whether something is in pinecone or not, as much as it stores what its state should be and tries to get it to that way. This will be fine in the case of items that are in pinecone but shouldn't be, as they will soon be removed, but if an item should be in pinecone (`pinecone_status == 'added'`), but for some reason isn't, then MySQL won't know that. The validator (new PR) should check that

align_data/db/models.py
```diff
@@ -43,6 +45,13 @@ class Summary(Base):
     article: Mapped["Article"] = relationship(back_populates="summaries")
 
 
+class PineconeStatus(enum.Enum):
+    absent = 1
+    pending_removal = 2
+    pending_addition = 3
+    added = 4
```
Thomas-Lemoine, line 52: by curiosity, why associate them with numbers? Oh, is it because the number is what really gets stored in the db, so it's cheaper to store, say, a number than a string? or just a convention, or maybe just shorter to write? 
To be sure I understand it right, the number should basically never be used by devs outside of that class definition right? They can always compare an enum using something like
```py
if PineconeStatus.absent == obj:
    do blabla
```
?

align_data/db/session.py
Thomas-Lemoine: Seems okay. If at some point, this set of articles to remove is very large, but the proportion of articles in that set that are actually in pinecone is really low, we can change it, but until then, seems ok

align_data/db/session.py
```diff
@@ -25,29 +25,43 @@ def make_session(auto_commit=False):
 def get_pinecone_articles(
     session: Session,
     force_update: bool = False,
+    statuses: List[PineconeStatus] = [PineconeStatus.pending_addition],
 ):
     return (
         session.query(Article)
-        .filter(or_(Article.pinecone_update_required.is_(True), force_update))
+        .filter(or_(Article.pinecone_status.in_(statuses), force_update))
```
Thomas-Lemoine, line 32: nice

align_data/db/models.py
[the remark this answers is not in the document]
Thomas-Lemoine: that makes sense

Thomas-Lemoine (approved):

align_data/db/models.py
mruwnik: exactly. The database also stores it as a string, so the only requirement is for it to be unique. The value could also just be the same as the key - it doesn't really matter

PR #182, opened 2023-09-03
Article checker

mruwnik:
Every 4h goes picks 100 articles that haven't been checked in the last 4 weeks and checks:
* whether there is new/better data
* whether the url is reachable

### DIFF
--- a/align_data/common/alignment_dataset.py
+++ b/align_data/common/alignment_dataset.py
@@ -5,10 +5,13 @@ import logging
 import time
 from dataclasses import dataclass, field, KW_ONLY
 from pathlib import Path
-from typing import List, Optional, Set, Iterable, Tuple, Generator
+from typing import Any, Dict, Iterable, List, Optional, Set, Tuple, Generator
+from sqlalchemy import select
+from sqlalchemy.exc import IntegrityError
+from sqlalchemy.orm import joinedload
 
 import pytz
-from sqlalchemy import select, Select, JSON
+from sqlalchemy import select, Select
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import joinedload, Session
 import jsonlines
@@ -21,6 +24,62 @@ from align_data.settings import ARTICLE_MAIN_KEYS
 from align_data.sources.utils import merge_dicts
 
 logger = logging.getLogger(__name__)
+
+
+
+def normalize_url(url: str | None) -> str | None:
+    if not url:
+        return url
+
+    # ending '/'
+    url = url.rstrip("/")
+
+    # Remove http and use https consistently
+    url = url.replace("http://", "https://")
+
+    # Remove www
+    url = url.replace("https://www.", "https://")
+
+    # Remove index.html or index.htm
+    url = re.sub(r'/index\.html?$', '', url)
+
+    # Convert youtu.be links to youtube.com
+    url = url.replace("https://youtu.be/", "https://youtube.com/watch?v=")
+
+    # Additional rules for mirror domains can be added here
+
+    # agisafetyfundamentals.com -> aisafetyfundamentals.com
+    url = url.replace("https://agisafetyfundamentals.com", "https://aisafetyfundamentals.com")
+
+    return url
+
+
+def normalize_text(text: str | None) -> str | None:
+    return (text or '').replace('\n', ' ').replace('\r', '').strip() or None
+
+
+def format_authors(authors: List[str]) -> str:
+    # TODO: Don't keep adding the same authors - come up with some way to reuse them
+    authors_str = ",".join(authors)
+    if len(authors_str) > 1024:
+        authors_str = ",".join(authors_str[:1024].split(",")[:-1])
+    return authors_str
+
+
+def article_dict(data, **kwargs) -> Dict[str, Any]:
+    data = merge_dicts(data, kwargs)
+
+    summaries = data.pop("summaries", [])
+    summary = data.pop("summary", None)
+
+    data['summaries'] = summaries + [summary] if summary else []
+    data['authors'] = format_authors(data.pop("authors", []))
+    data['title'] = normalize_text(data.get('title'))
+
+    return dict(
+        meta={k: v for k, v in data.items() if k not in ARTICLE_MAIN_KEYS and v is not None},
+        **{k: v for k, v in data.items() if k in ARTICLE_MAIN_KEYS},
+    )
 
 
 @dataclass
@@ -64,28 +123,10 @@ class AlignmentDataset:
     def __str__(self) -> str:
         return self.name
 
-    def _add_authors(self, article: Article, authors: List[str]) -> Article:
-        # TODO: Don't keep adding the same authors - come up with some way to reuse them
-        article.authors = ",".join(authors)
-        if len(article.authors) > 1024:
-            article.authors = ",".join(article.authors[:1024].split(",")[:-1])
-        return article
-
     def make_data_entry(self, data, **kwargs) -> Article:
-        data = merge_dicts(data, kwargs)
-
-        summaries = data.pop("summaries", [])
-        summary = data.pop("summary", None)
-        summaries += [summary] if summary else []
-
-        authors = data.pop("authors", [])
-        data['title'] = (data.get('title') or '').replace('\n', ' ').replace('\r', '') or None
-
-        article = Article(
-            meta={k: v for k, v in data.items() if k not in ARTICLE_MAIN_KEYS and v is not None},
-            **{k: v for k, v in data.items() if k in ARTICLE_MAIN_KEYS},
-        )
-        self._add_authors(article, authors)
+        data = article_dict(data, **kwargs)
+        summaries = data.pop('summaries', [])
+        article = Article(**data)
         article.summaries += [Summary(text=summary, source=self.name) for summary in summaries]
         return article
 
@@ -109,7 +150,7 @@ class AlignmentDataset:
             query = self._query_items.options(joinedload(Article.summaries))
             if sort_by is not None:
                 query = query.order_by(sort_by)
-            
+
             result = session.scalars(query)
             for article in result.unique(): # removes duplicates
                 yield article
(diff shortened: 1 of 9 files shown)


align_data/common/alignment_dataset.py
```diff
@@ -23,6 +23,62 @@
 logger = logging.getLogger(__name__)
 
 
+
+def normalize_url(url: str | None) -> str | None:
```
Thomas-Lemoine, line 27: It this is here, shouldn't normalize_urls also be here?
```py
def normalize_urls(urls: Iterable[str]) -> Set[str]:
    return {normalize_url(url) for url in urls}
```

mruwnik: dunno. I only wanted to extract what I actually need. I was also considering to move these function to a separate module or something, rather than having them here. What would you prefer?

align_data/sources/validate.py
```diff
@@ -0,0 +1,84 @@
+import logging
```
Thomas-Lemoine, line 1: I don't have a strong opinion on it, but would it make more sense to have this file in align_data/db, since it's used to validate db objects, rather than any source in particular (afaict)

mruwnik: sort of. It also makes use of the parsers, as it fetches data from the appropriate source etc., so dunno :/

align_data/sources/validate.py
```diff
@@ -0,0 +1,84 @@
...
+    if field == 'title' and normalize_text(article.title) == normalize_text(value):
+        # If there are slight differences in the titles (e.g. punctuation), assume the
+        # database version is more correct
+        return
+    if field == 'meta':
+        article.meta = article.meta or {}
+        for k, v in value.items():
+            meta_val = article.meta.get(k)
+            if not meta_val or v > meta_val:
+                article.meta[k] = v
```
Thomas-Lemoine, line 33: If the values can't be compared, does it throw an error?

mruwnik: yes. I suppose metadata keys can be ignored in those cases...

align_data/sources/validate.py
```diff
@@ -0,0 +1,84 @@
...
+
+    article_val = getattr(article, field, None)
+    # Assume that if the provided value is larger (or later, in the case of dates), then it's
+    # better. This might very well not hold, but it seems like a decent heuristic?
+    if not article_val:
+        setattr(article, field, value)
+    elif isinstance(value, datetime) and value > article_val:
+        setattr(article, field, value)
+    elif isinstance(value, str) and len(normalize_text(value) or '') > len(normalize_text(article_val) or ''):
+        setattr(article, field, normalize_text(value))
```
Thomas-Lemoine, line 44: the > heuristic seems fine since it will only do anything in the case of comparables like dates and numbers and idk what else, but imo, in other cases it would always update it? Like, if the text changed it would always be updated, rather than only updated when the length of the text grows, which seems like what's going on here. Lmk if I'm wrong

mruwnik: Text changes seem to be:
* slight edits for stylistic reasons etc., which often have a note attached to say when the edit happened - I'm hoping these will result in slightly larger texts
* things being ammended - this should result in a larger article
* things being removed, or the article disappearing with only an error message displayed

I didn't want to just overwrite the text if it changed, because the changes might not be for the better. So this seemed like a heuristic that might be better than that. Albeit not much..

align_data/sources/validate.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def check_article(article: Article) -> Article:
```
Thomas-Lemoine, line 47: It seems like this would be worth either changing the name or creating a docstring, since the name "check_article" doesn't seem to imply much of what it does. 
Seems like:
1) it tries to get the source's url
2) tries to fetch its contents, which should only ever work if the url is from a domain that has a parser implemented specifically for it. 
3) if it gets anything and it doesn't return an error (indicating a parser exists for this url and the parse was successful), it tries to update the fields of the corresponding article. 
4) if the url is not reachable, it is noted as such. 
5) the date_checked is updated.

Maybe "update_article" would be a sufficiently good name for that? 
and `update_article` goes through the fields and runs `update_article_fields`, which makes sense?

align_data/sources/validate.py
```diff
@@ -0,0 +1,84 @@
...
+    if 400 <= fetch(article.url).status_code < 500:
+        logger.info('Could not get url for %s', article)
+        article.status = 'Unreachable url'
+
+    article.date_checked = datetime.utcnow()
+
+    return article
+
+
+def check_articles(sources: List[str], batch_size=100):
```
Thomas-Lemoine, line 68: similarly as above, maybe `update_articles`? since that's what it does, updates them to the newest contents if possible.

align_data/common/alignment_dataset.py
Thomas-Lemoine: Ah, the reason you need it is that you want to be able to use normalize_url and others without using AlignmentDataset at all, correct? Since `normalize_urls` is never used without having access to a dataset obj, it can stay in AlignmentDataset?

Thomas-Lemoine: I have a small preference for keeping it in the `align_data/common/`. What do you think about some format.py file in common/ that deals with common ways to reformat different objects throughout the code, like formatting url strings to a standard-ish format, formatting authors list to a standard authors format, format an entry's contents dict into some standard-but-flexible article_data_dict format which make_data_entry expects, etc.?

align_data/sources/validate.py
Thomas-Lemoine: What are you thinking would be situations where the more recent changes are worse? If I'm understanding it right, "contents" will only be successful (no errors, non empty) when the url corresponds to a parsers-friendly url, ie a url that was originally fetched using the parsers, right? In that case, the newly fetched `contents` will be created through the same method as the original contents of the article, but will have (as the only difference) more recent data

Thomas-Lemoine: Actually, it's meant to also work with any pdf and epub. Is that the reason why the contents may be of lesser quality? if the url is not the thing we used to get the pdf in the first place, maybe this generic pdf parser will be worse than a specialized one. Seems tricky.

Thomas-Lemoine:
If my understanding is right, validation's purpose is:
1) for urls that have parsers, link to a pdf or link to an epub, it updates them somewhat.
2) regardless of 1), check that the url works, and update the article's status accordingly.
3) update the date_checked

It seems to me that 1) and 2+3 are pretty different tasks? Would it maybe make more sense to have the validator file have the following purpose:
It goes through all sources, and presses on some of the links to see if they work. if they don't, add article.status of "Unreachable url", but update the date_checked in any case. It could have a name like "check_article_url", to indicate that it's only really checking that the url hasn't closed or something, since it was fetched originally.

And then, we could just refetch parsers + pdfs + epubs more often?

align_data/sources/validate.py
mruwnik: exactly. Plus some websites return an error in HTML, but with a 200 code, or other such madness, so you can't really trust them :/
Though it might be better to just always replace things if possible...

mruwnik:
Regular fetching more often won't work, as the regular fetchers only get new data -  they ignore things they've already seen. The basic assumption is that things won't change that often, but it's worth checking every now and then if there are updates. It would be possible to always recheck all urls, but that would take a LOT more time than the current process of only looking for new items.
Ideally, I'd like to move away from there being multiple ways of fetching data to having the list of parsers also contain things that are handled in other places (e.g. have parsers for LW) in there. This is also a mechanism that would sorta-automatically fix any articles that previously couldn't be parsed, e.g. because they didn't have a domain handler.

Thomas-Lemoine:
Hmmm, I see. would something like
- `regular fetching ever X days`
- `fetching with --rebuild, or with an empty self._outputted_items essentially, at longer intervals`
work? my sense of how much slower it is to fetch all vs some is pretty weak. additionally, is fetching an individual article slower than updating it? I suppose it seems pretty similar

align_data/sources/validate.py
```diff
@@ -66,6 +71,7 @@ def check_article(article: Article) -> Article:
 
 
 def check_articles(sources: List[str], batch_size=100):
```
Thomas-Lemoine, line 73: this is really minor, but for consistency the first and second function should have the same name, except one plural, imo (ie update_article and update_articles or check_article and check_articles, since one just calls the other on a set of lots of articles

align_data/common/formatters.py
```diff
@@ -0,0 +1,60 @@
+import re
```
Thomas-Lemoine, line 1: nice

mruwnik:
To a certain extent. It would be fine with smaller datasets, which can take a few seconds or minutes to run, but the larger ones (e.g. LW, arxiv) can take hours, which is why I wanted to have a hundred or so checked every few hours.
An issue with the `--rebuild` option is that datasets would always have to do upserts, rather than inserts. Though that actually might be a good thing?

align_data/sources/validate.py
```diff
@@ -0,0 +1,90 @@
...
+    if field == 'meta':
+        article.meta = article.meta or {}
+        for k, v in value.items():
+            meta_val = article.meta.get(k)
+            try:
+                if not meta_val or v > meta_val:
+                    article.meta[k] = v
+            except Exception as e:
+                # Ignore exceptions here - the metadata isn't that important
+                logger.info('Error checking metadata value: %s', value)
```
Thomas-Lemoine, line 37: NIT - I'm imagining seeing the error logged and having no idea what article it's for; would adding the article url, or something, be beneficial

Thomas-Lemoine:
> To a certain extent. It would be fine with smaller datasets, which can take a few seconds or minutes to run, but the larger ones (e.g. LW, arxiv) can take hours, which is why I wanted to have a hundred or so checked every few hours. 

I see. I'm completely on board for doing manageable batches at a time for updating. Suppose we call `updating_article` the act of making a single article more up-to-date, which currently only works for parsers + pdfs + epubs, and we call `validating_url` the act of checking the status code of a fetch of a url and seeing if it works (and modifying `.status` accordingly), as well as updating the `date_checked`. Then, I have a slight preference for separating them since, as far as I understand, this would save some (maybe significant? unsure) fetching time, and give us control over what we do. This is an example of what I have in mind:
- every X days, validate all the urls, or every 4 hours, validate batches of urls
- separately, update articles from certain sources at some frequency. The idea is that, then, we can decide if one source has a parser-focused updating method, and another has a (perhaps much rarer, but can be automatic) refetching of all the sources, `--rebuild` style. Like, for for each source s_i we decide if it is
1) Never updated
2) updated as one of (parsables + pdfs + epubs), either as batches every X hours or as a whole every N days
3) updated with `refetching + --rebuild` or similar, every N_i days

This way, datasets that get very little marginal value from getting updated (like arxivs; there are a lot of them and very few are being meaningfully modified) can be updated very rarely or never, whereas datasets that get a lot of value from updates (aisafety.info especially) can be updated every day or so. Finally, datasets that can be updated url-by-url, which gives the advantage of being updateable only using mysql's url data (parsable urls), can be updated in small batches, which spreads out the updates.

Maybe this makes it needlessly more complex, though. Also kind of points to "date_updated" for `update_article` being implied by the existence of "date_checked" for `check_article`, but I'm not sure; an advantage of combining those two tasks is that those two dates are combined.


> An issue with the `--rebuild` option is that datasets would always have to do upserts, rather than inserts. Though that actually might be a good thing?
I don't know for sure, but yeah in my mind upserts are always better, I think? We keep old information if it's not replaced by new, but any new information is assumed to be improved or better than the original. One potential risk is that some url has its text replaced with "This article is no longer available" or something, which would be a bummer.

mruwnik:
I considered adding special handling for that, but decided against it - the entry won't be deleted, it'll just have its status changed. So it should be possible to then just check items with that error and remove the error status, at which point it should be reindexed. Of course, if this starts being an issue it'll have to be addressed.
There isn't a mysql trigger, but the main updater task will add unadded items to pinecone, and remove items that are in it but have error statuses

mruwnik:
(I wrote this comment right away, but didn't post it...)
I considered adding special handling for that, but decided against it - the entry won't be deleted, it'll just have its status changed. So it should be possible to then just check items with that error and remove the error status, at which point it should be reindexed. Of course, if this starts being an issue it'll have to be addressed.
There isn't a mysql trigger, but the main updater task will add unadded items to pinecone, and remove items that are in it but have error statuses

PR #189, opened 2023-09-13
added some configuration instructions to readme

Thomas-Lemoine:
some things left:
- I don't exactly understand how airtable stuff works, and there's no instructions for making that work in the readme
- `--rebuild` was removed, not sure if we ought to remove it from the readme or make a similar thing available in some way. like either dropping the articles from a source before going through them or upserting for each or wtv.

I didn't test these: agisf arxiv indices special_docs youtube, since they were either very slow or didn't work at the time

Some things I may have changed that were best left unchanged, if so lmk

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,53 +64,175 @@ Additional keys may be available depending on the source document.
 
 ## Development Environment
 
-To set up the development environment, run the following steps:
+### 1. Clone the repository:
 
-```bash
+```sh
 git clone https://github.com/StampyAI/alignment-research-dataset
 cd alignment-research-dataset
+```
+
+### 2. Set up Environment Variables:
+
+Duplicate the provided `.env.example` to create your environment configuration:
+
+```sh
+cp .env.example .env
+```
+
+This `.env` file contains placeholders for several configuration options. Further details about how to configure them are in the [Configuration section](#configuration).
+
+### 3. Install Dependencies:
+
+```sh
 pip install -r requirements.txt
 ```
 
-### Database
-
-You'll also have to set up a MySQL database. To do so with Docker, you can run `./local_db.sh` which should spin up a container
-with the database initialised.
-
-### CLI options
-
-The available CLI options are list, fetch, fetch-all, and count-tokens.
-
-To get a list of all available datasets:
+**Optional:** For testing purposes, you can also install testing dependencies:
 
 ```sh
-python main.py list
+pip install -r requirements-test.txt
 ```
 
-To fetch a specific dataset, replace [DATASET_NAME] with the name of the dataset you want to fetch. The optional `--rebuild` parameter allows you to remove the previous build before running, scraping everything from scratch. Otherwise, only the new files will be scraped.
+### 4. Database Setup:
+
+Initialize a MySQL database. To do so with [Docker](https://docs.docker.com/get-docker/), and spin up a container with the database initialised, run the following:
 
 ```sh
-python main.py fetch [DATASET_NAME] --rebuild
+./local_db.sh
 ```
 
-The command to fetch all datasets is below. Again, the optional `--rebuild` parameter allows you to scrape everything from scratch.
+## Configuration
 
+Various subcomponents in this project rely on external services, so need credentials set. This is done via environment variables. The file `.env` is the central location for these settings.
+
+### Logging
+
+The log level can be configured with the `LOG_LEVEL` environment variable. The default level is 'WARNING'.
+
+### Coda
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -64,146 +64,206 @@ Additional keys may be available depending on the source document.
...
+### AlignmentDataset class
 
-    python main.py update_metadata <input spreadsheet id> <input sheet name> <output spreadsheet id>
+This is the main workhorse for processing datasets. The basic idea is that it provided a list of items to be processed, and after processing a given item, appends it to the appropriate jsonl file, where each line of the file is a JSON object with all the data. The `AlignmentDataset` class has various methods that can be implemented to handle various cases. A few assumptions are made as to the data it will use, i.e.:
```
mruwnik, line 240: this should be overhauled... :D  Though mainly by removing the jsonl info and adding that it saves it to MySQL

README.md
```diff
@@ -64,146 +64,206 @@ Additional keys may be available depending on the source document.
...
 
-which for the current documents would be:
+* `self.data_path` is where data will be written to and read from - by default it's the `data/` directory
+* `self.raw_data_path` is where downloaded files etc. should go - by default it's the `data/raw` directory
+* `self.files_path` is where data to be processed is expected to be. This is used e.g. when a collection of html files are to be processed
```
mruwnik, line 244: these also aren't really used now

README.md
```diff
@@ -64,146 +64,206 @@ Additional keys may be available depending on the source document.
...
 
-    python main.py update_metadata 1pgG3HzercOhf4gniaqp3tBc3uvZnHpPhXErwHcthmbI special_docs.csv 1l3azVJVukGAvZPgg0GyeqiaQe8bEMZvycBJaA8cRXf4
+The `AlignmentDataset` is a dataclass, so it has a couple of settings that control it:
 
-#### Google API
+* `name` - this is a string that identifies the dataset, i.e. 'lesswrong'
+* `done_key` - used to check if a given item has already been processed. This is a key in the JSON object that gets written to the output file - any subsequent entries with the same value for that key will be skipped
```
mruwnik, line 249: this should be removed

README.md
```diff
@@ -64,146 +64,206 @@ Additional keys may be available depending on the source document.
...
- * Enable the Google Sheets API for your project at https://console.cloud.google.com/apis/api/sheets.googleapis.com/metrics?project=<your project id>
- * Enable the Google Drive API for your project at https://console.cloud.google.com/apis/api/drive.googleapis.com/metrics?project=<your project id>
- An alternative to this step is that when running the program without these enabled, an exception will be raised telling you how to enable it - you can then just open the link in the exception message
-4. Navigate to the "Credentials" section
-5. Click on "Create Credentials" and select "Service Account"
-6. Fill in the required information for the service account
-7. On the "Create key" page, select the JSON key type and click "Create"
-8. The JSON file containing your credentials will be downloaded -> save as credentials.json in the folder from which you're running the code
+1. `self.setup()` - any instance level initialization stuff should go here, e.g. fetching zip files with data
+2. `self._load_outputted_items()` - go through `self.jsonl_path` and construct a set of the `self.done_key` values of each item - this is used to skip items that have already been processed
```
mruwnik, line 255: this should be changed to say that it gets the urls of previously processed items from the database

README.md
Thomas-Lemoine: ah, good catch

README.md
Thomas-Lemoine: they seem only to be used by datasets.py and by agentmodels.py

README.md
Thomas-Lemoine: Do I just keep "* `done_key` - used to check if a given item has already been processed."?

README.md
Thomas-Lemoine: Is my understanding wrong that this still uses done_key and not strictly urls? If not, how about this:
```
2. `self._load_outputted_items()` - go through articles in the database, loads the value of their `self.done_key`, and skips items that have a `self.done_key` already present in the db
```

Thomas-Lemoine: """
2. `self._load_outputted_items()` - go through articles in the database, loads the value of their `self.done_key`, outputs a simplefied version of these strings. The purpose for this is to then compare new items against these _outputted_items, and skip items that are already in the database
"""
is a bit more explicit

README.md
mruwnik: yup. Maybe with a note that the default is to just check the `url` field

mruwnik (approved):

PR #190, opened 2023-09-13
add debug logging for logging that takes up a lot of space

Thomas-Lemoine:
avoids logger.info-ing all items in a source, and instead logs some constant amount of time per source, rather than proportional to its length, with the default logger level (info). if you're debugging you can set that to "DEBUG" for more info.

### DIFF
--- a/align_data/sources/blogs/gwern_blog.py
+++ b/align_data/sources/blogs/gwern_blog.py
@@ -71,7 +71,7 @@ class GwernBlog(HTMLDataset):
         return dict(filter(None, map(extract, header.splitlines())))
 
     def _get_article(self, url):
-        logger.info("Fetching {}".format(url))
+        logger.debug(f"Fetching {url}")
         return requests.get(url, allow_redirects=True)
 
     @staticmethod
--- a/main.py
+++ b/main.py
@@ -60,7 +60,7 @@ class AlignmentDataset:
         """
         names = [name for name in ALL_DATASETS if name not in skip]
         for name in names:
-            print(name)
+            logger.debug(name)
             self.fetch(name)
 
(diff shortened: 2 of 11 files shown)


align_data/sources/blogs/gwern_blog.py
```diff
@@ -71,7 +71,7 @@ def extract(item):
         return dict(filter(None, map(extract, header.splitlines())))
 
     def _get_article(self, url):
-        logger.info("Fetching {}".format(url))
+        logger.debug(f"Fetching {url}")
```
mruwnik, line 74: strictly speaking, when logging you should use params, like `logger.debug("Fetching %s", url)`. It's fine here, but can potentially be a problem if casting `url` to a string is a slow or costly operation - with fstrings it's always done, but when provided as an argument it only does it when needed (i.e. log level == 'debug'). Again - here it's fine, but it's worth bearing in mind

main.py
```diff
@@ -74,7 +74,7 @@ def generate_jsonl_files(self, *names):
         assert not missing, f"{missing} are not valid dataset names"
         for name in names:
             dataset = get_dataset(name)
-            print(dataset.to_jsonl())
+            logger.info(dataset.to_jsonl())
```
mruwnik, line 77: `logger.info("%s", dataset.to_jsonl())`, or `logger.info(str(dataset.to_jsonl()))`. `logger.info` expects a string as its first argument

align_data/sources/blogs/gwern_blog.py
Thomas-Lemoine: Oh, I see, I hadn't thought of that

PR #191, opened 2023-10-15
Fix moderation batching

mruwnik:


align_data/embeddings/embedding_utils.py
```diff
@@ -93,11 +93,31 @@ def _single_batch_moderation_check(batch: List[str]) -> List[ModerationInfoType]
...
-    """Batch moderation checks on list of texts."""
+def moderation_check(texts: List[str], max_batch_size: int = 4096, tokens_counter: Callable[[str], int] = len) -> List[ModerationInfoType]:
+    """Batch moderation checks on list of texts.
+
+    :param List[str] texts: the texts to be checked
+    :param int max_batch_size: the max size in tokens for a single batch
+    :param Callable[[str], int] tokens_counter: the function used to count tokens
+    """
+    # A very ugly loop that will split the `texts` into smaller batches so that the
+    # total sum of tokens in each batch will not exceed `max_batch_size`
```
mruwnik, line 104: as the comment says. This is very ugly. I couldn't think of anything better that wouldn't be a lot more complicated

align_data/embeddings/embedding_utils.py
```diff
@@ -93,11 +93,31 @@ def _single_batch_moderation_check(batch: List[str]) -> List[ModerationInfoType]
     return openai.Moderation.create(input=batch)["results"]
 
 
-def moderation_check(texts: List[str], max_texts_num: int = 32) -> List[ModerationInfoType]:
-    """Batch moderation checks on list of texts."""
+def moderation_check(texts: List[str], max_batch_size: int = 4096, tokens_counter: Callable[[str], int] = len) -> List[ModerationInfoType]:
```
mruwnik, line 96: this uses `len` to calculate token usage, which will overestimate by something like 2-3 times. It's also used elsewhere here, from what I can see, so I left it for now

align_data/embeddings/embedding_utils.py
mruwnik: yup, both are terrible :P I wanted something that would do it in say around 2-3 lines. This offends my sensibilities. But it's better than any alternatives :/

PR #198, opened 2024-04-21
GreaterWrong: Fix stuck scraper iteration

jbeshir:
Unclear what exactly causes this, but on some incremental runs the scraper gets the same single item over and over again, failing to progress.

This looks like sometimes setting the `after` to the `postedAt` of a post can return that post, and if this happens for the last item it will loop forever.

This PR adds a check to catch this case and return, as well as a more general check that throws if the date fails to advance (so we bail out rather than spamming duplicate requests if we do get stuck).

This also makes the cooldown setting for the AF/LW/EA scrapers to be effective (the field used everywhere else is `COOLDOWN` rather than `COOLDOWN_TIME`).

### DIFF
--- a/align_data/sources/greaterwrong/greaterwrong.py
+++ b/align_data/sources/greaterwrong/greaterwrong.py
@@ -182,16 +182,26 @@ class GreaterWrong(AlignmentDataset):
     def items_list(self):
         next_date = self.last_date_published
         logger.info("Starting from %s", next_date)
+        last_item = None
         while next_date:
             posts = self.fetch_posts(self.make_query(next_date))
             if not posts["results"]:
+                return
+
+            # If the only item we find was the one we advanced our iterator to, we're done
+            if len(posts["results"]) == 1 and last_item and posts["results"][0]["pageUrl"] == last_item["pageUrl"]:
                 return
 
             for post in posts["results"]:
                 if post["htmlBody"] and self.tags_ok(post):
                     yield post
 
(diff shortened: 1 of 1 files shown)


align_data/sources/greaterwrong/greaterwrong.py
```diff
@@ -67,7 +67,7 @@ class GreaterWrong(AlignmentDataset):
     """Whether alignment forum posts should be returned"""
 
     limit = 50
-    COOLDOWN_TIME: float = 0.5
+    COOLDOWN = 0.5
```
mruwnik, line 70: nice catch

align_data/sources/greaterwrong/greaterwrong.py
```diff
@@ -182,16 +182,26 @@ def last_date_published(self) -> str:
...
         next_date = self.last_date_published
         logger.info("Starting from %s", next_date)
+        last_item = None
         while next_date:
             posts = self.fetch_posts(self.make_query(next_date))
             if not posts["results"]:
                 return
 
+            # If the only item we find was the one we advanced our iterator to, we're done
+            if len(posts["results"]) == 1 and last_item and posts["results"][0]["pageUrl"] == last_item["pageUrl"]:
```
mruwnik, line 192: would adding a small delta to `next_date` fix things? e.g. 1s?

mruwnik (approved):

PR #200, opened 2024-04-21
Airtable: Take base and table IDs from environment

jbeshir:
This avoids an issue where Airtable doesn't allow you to create API keys with access to bases you don't own, meaning only the owner of the airtable base added to the dataset can scrape it.


.env.example
```diff
@@ -3,7 +3,10 @@ MIN_CONFIDENCE="0.5"
 
 CODA_TOKEN=""
 YOUTUBE_API_KEY=""
+
 AIRTABLE_API_KEY=""
+AIRTABLE_BASE_ID=""
```
mruwnik, line 8: could you change this to `AGISF_AIRTABLE_BASE_ID` etc.? I'm a bit worried about what would happen if more airtable sources are added

mruwnik (changes requested):

.env.example
jbeshir: Changed, and also moved where they're used up a level so they're used when configuring the dataset, where the hardcoded IDs used to be, so the dataset-specific stuff stays separate from the general airtable stuff.

mruwnik: nice! I was just about to ask you to do that :D

mruwnik (approved):

PR #201, opened 2024-04-21
Add Dockerfile to run the scraper in a container

jbeshir:
Adds a Dockerfile and makes a couple of small changes to make running in a serverless container environment easier:

- Takes a full DB connection URI rather than pieces of one, so it can use a UNIX domain socket to connect to the DB in environments which provide one of those. As implemented here, this requires updating the secrets feeding GitHub workflows to provide this instead of the individual pieces.
- Moves GCP credentials to a subdirectory rather than the root of the repository, so they can be mounted into the container as a volume from a secret.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -123,4 +123,6 @@ carado.moe/
 *.epub
 
 secrets/
+credentials.json
+
 data/raw/
--- a/.gitignore
+++ b/.gitignore
@@ -122,5 +122,5 @@ carado.moe/
 !requirements.txt
 *.epub
 
-credentials.json
+secrets/
 data/raw/
--- a/align_data/settings.py
+++ b/align_data/settings.py
@@ -34,12 +34,7 @@ YOUTUBE_API_KEY = os.environ.get("YOUTUBE_API_KEY")
 AIRTABLE_API_KEY = os.environ.get("AIRTABLE_API_KEY")
 
 ### MYSQL ###
-user = os.environ.get("ARD_DB_USER", "user")
-password = os.environ.get("ARD_DB_PASSWORD", "we all live in a yellow submarine")
-host = os.environ.get("ARD_DB_HOST", "127.0.0.1")
-port = os.environ.get("ARD_DB_PORT", "3306")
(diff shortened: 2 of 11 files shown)


align_data/settings.py
```diff
@@ -34,12 +34,7 @@
 AIRTABLE_API_KEY = os.environ.get("AIRTABLE_API_KEY")
 
 ### MYSQL ###
-user = os.environ.get("ARD_DB_USER", "user")
-password = os.environ.get("ARD_DB_PASSWORD", "we all live in a yellow submarine")
-host = os.environ.get("ARD_DB_HOST", "127.0.0.1")
-port = os.environ.get("ARD_DB_PORT", "3306")
-db_name = os.environ.get("ARD_DB_NAME", "alignment_research_dataset")
-DB_CONNECTION_URI = f"mysql+mysqlconnector://{user}:{password}@{host}:{port}/{db_name}"
+DB_CONNECTION_URI = os.environ.get("ARD_DB_CONNECTION_URI", "mysql+mysqlconnector://user:we all live in a yellow submarine@127.0.0.1:3306/alignment_research_dataset")
```
mruwnik, line 37: why not have support for both? i.e. something like:
```
if not (DB_CONNECTION_URI := os.environ.get("ARD_DB_CONNECTION_URI"):
    user = os.environ.get("ARD_DB_USER", "user")
    password = os.environ.get("ARD_DB_PASSWORD", "we all live in a yellow submarine")
    DB_CONNECTION_URI = f"mysql+mysqlconnector://{user}:{password}@{host}:{port}/{db_name}"
```
I'd like to keep the possibility of overriding single parameters, as it makes it easier to do quick changes

.gitignore
```diff
@@ -122,5 +122,5 @@ carado.moe/
 !requirements.txt
 *.epub
 
-credentials.json
```
mruwnik, line 125: keep em both

mruwnik (changes requested):

align_data/settings.py
jbeshir: Makes sense, done.

mruwnik (approved):
nice

PR #204, opened 2025-02-19
Update Pinecone fetching

LeMurphant:
Updates were failing, they now work when I run the script locally.

There are probably better ways to fix this, feel free to suggest them.

### DIFF
--- a/align_data/sources/articles/html.py
+++ b/align_data/sources/articles/html.py
@@ -41,7 +41,27 @@ def fetch(
 
     This function is to have a single place to manage headers etc.
     """
-    return getattr(requests, method)(url, allow_redirects=True, headers=headers)
+    try:
+        session = requests.Session()
+        adapter = requests.adapters.HTTPAdapter(max_retries=5)
+        session.mount('http://', adapter)
+        session.mount('https://', adapter)
+        
+        return session.request(
+            method=method,
+            url=url, 
+            allow_redirects=True, 
+            headers=headers
+        )
+    except requests.exceptions.TooManyRedirects:
+        # Create a response-like object with error status
+        mock_response = requests.Response()
+        mock_response.status_code = 429
+        mock_response._content = b"Too many redirects"
+        mock_response.headers = {"Content-Type": "text/plain"}
+        return mock_response
+    finally:
+        session.close()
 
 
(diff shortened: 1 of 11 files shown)


LeMurphant:
Patched many of the tests. Again, there was probably a more principled way to do this.

align_data/sources/articles/html.py
```diff
@@ -41,7 +41,27 @@ def fetch(
 
     This function is to have a single place to manage headers etc.
     """
-    return getattr(requests, method)(url, allow_redirects=True, headers=headers)
+    try:
+        session = requests.Session()
```
Aprillion, line 45: not sure what this fuction needs to achieve and what was the bug that the new code is solving, but I imagine it would fetch more than 1 url per minute .. and a Session abstraction was indended for handling (ahem) a session with all those requests in 1 session, not separate session per request => I would imagine creating the Session outside of the fetch function if that's possible

align_data/sources/articles/html.py
```diff
@@ -41,7 +41,27 @@ def fetch(
...
+            allow_redirects=True, 
+            headers=headers
+        )
+    except requests.exceptions.TooManyRedirects:
+        # Create a response-like object with error status
+        mock_response = requests.Response()
+        mock_response.status_code = 429
+        mock_response._content = b"Too many redirects"
+        mock_response.headers = {"Content-Type": "text/plain"}
+        return mock_response
```
Aprillion, line 62: would following work? when does the exception happen - after a real response is available or do we need to handle some internal exception that is not a real 429 response?
```
    except requests.exceptions.TooManyRedirects as e:
        return e.response
```

Aprillion (approved):

align_data/sources/articles/html.py
LeMurphant: This solves an apparent infinite redirect issue I had when initially running the script
[url_loop.txt](https://github.com/user-attachments/files/18893027/url_loop.txt)
I wanted to catch this error. I also added a max number of retries because the default with this method could loop for a while
I agree the solution is probably too complicated, I'll try the simple try-except loop, I just have to figure out how to re-create the error.

LeMurphant: The simple try-catch worked fine

Aprillion (approved):

Issue #208, opened 2025-04-23
Testing for LW tags does not work


LeMurphant:
When running `python main.py pinecone_update lesswrong` I notice that *all* LW posts get imported, not just the AI ones.

It looks like the structure of https://www.lesswrong.com/tag/ai has changed, and since it can't fetch tags the `self.ai_tags` set is empty, which causes all articles to be accepted. This means that updating LW takes a very long time.

LeMurphant:
While we are there, it might be worth initially fetching the number of new articles that will be integrated in a specific run, and printing a %, since my current run has been going over 1 hour, and I don't know if it will finish in 1 hour or 10 hours.

PR #210, opened 2025-05-05
Fix tag fetching

LeMurphant:
Fix https://github.com/StampyAI/alignment-research-dataset/issues/208

Ended up using fixed tags to fetch on LW and EAF.

I also implemented display of progress, since running the pinecone update can take a while.

I'd be interested in feedback on cleanup.py, and whether this was the best way to do it.

@mruwnik I was a bit confused about the --rebuild option in the README, since it is never implemented, but maybe maps to --force_update by default?

### DIFF
--- a/align_data/embeddings/pinecone/pinecone_db_handler.py
+++ b/align_data/embeddings/pinecone/pinecone_db_handler.py
@@ -6,9 +6,11 @@ from typing import List, Tuple
 # Try to handle both new and old Pinecone API patterns
 try:
     from pinecone import Pinecone
+
     USING_NEW_API = True
 except (ImportError, AttributeError):
     import pinecone
+
     USING_NEW_API = False
 
 from urllib3.exceptions import ProtocolError
@@ -188,22 +190,28 @@ class PineconeDB:
             result = []
             for match in matches:
                 # Handle both dictionary and object formats
-                match_id = match.get("id") if isinstance(match, dict) else getattr(match, "id", None)
-                match_score = match.get("score") if isinstance(match, dict) else getattr(match, "score", None)
-                
+                match_id = (
+                    match.get("id") if isinstance(match, dict) else getattr(match, "id", None)
+                )
+                match_score = (
+                    match.get("score") if isinstance(match, dict) else getattr(match, "score", None)
+                )
+
                 # Handle metadata in both formats
                 if isinstance(match, dict):
                     metadata = match.get("metadata", {})
                 else:
                     metadata = getattr(match, "metadata", {})
-                
+
                 # Create a dictionary representation for consistent interface
-                result.append({
-                    "id": match_id,
-                    "score": match_score,
-                    "metadata": metadata  # We'll use the metadata directly instead of wrapping in PineconeMetadata
-                })
-                
+                result.append(
+                    {
+                        "id": match_id,
+                        "score": match_score,
+                        "metadata": metadata,  # We'll use the metadata directly instead of wrapping in PineconeMetadata
+                    }
+                )
+
             return result
         except Exception as e:
             logger.error(f"Failed to parse query response: {e}")
--- a/align_data/embeddings/pinecone/update_pinecone.py
+++ b/align_data/embeddings/pinecone/update_pinecone.py
@@ -183,6 +183,17 @@ class PineconeUpdater(PineconeAction):
 
         :param custom_sources: List of sources to update.
         """
+        # Check if we need to clean up LessWrong posts
+        if 'lesswrong' in custom_sources:
+            try:
+                from align_data.sources.greaterwrong.cleanup import cleanup_lesswrong_posts
+                logger.info('Cleaning up non-AI LessWrong posts')
+                cleanup_lesswrong_posts(dry_run=False, source='lesswrong')
+            except ImportError:
+                logger.warning('Could not import cleanup_lesswrong_posts - skipping cleanup')
+            except Exception as e:
+                logger.error(f'Error cleaning up LessWrong posts: {e}')
+                
         logger.info('Adding pinecone entries for %s', custom_sources)
         self.adder.update(custom_sources, force_update)
 
--- a/align_data/embeddings/pinecone/update_pinecone.py
+++ b/align_data/embeddings/pinecone/update_pinecone.py
@@ -47,16 +47,43 @@ class PineconeAction:
         """
         with make_session() as session:
             articles_to_update = self._articles_by_source(session, custom_sources, force_update)
-            logger.info('Processing %s items', articles_to_update.count())
+            total_articles = articles_to_update.count()
+            
+            logger.info('Processing %s items from %s', total_articles, custom_sources)
+            
+            total_processed = 0
+            update_interval = 10
+            
             for batch in self.batch_entries(articles_to_update):
                 self.save_batch(session, batch)
+                total_processed += len(batch)
+                
+                if total_processed % update_interval == 0 or total_processed == total_articles:
+                    percentage = (total_processed / total_articles) * 100 if total_articles > 0 else 0
+                    logger.info('Progress: %.1f%% (%d/%d)', percentage, total_processed, total_articles)
+            
+            logger.info('Completed processing %s items', total_processed)
 
     def update_articles_by_ids(self, hash_ids: List[int], force_update: bool = False):
         """Update the Pinecone entries of specific articles based on their hash_ids."""
         with make_session() as session:
             articles_to_update = self._articles_by_id(session, hash_ids, force_update)
+            total_articles = articles_to_update.count()
+            
+            logger.info('Processing %s items by ID', total_articles)
+            
+            total_processed = 0
+            update_interval = 10 
+            
             for batch in self.batch_entries(articles_to_update):
                 self.save_batch(session, batch)
+                total_processed += len(batch)
+                
+                if total_processed % update_interval == 0 or total_processed == total_articles:
+                    percentage = (total_processed / total_articles) * 100 if total_articles > 0 else 0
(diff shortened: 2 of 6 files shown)


LeMurphant:
Inserting works ok, but there are issues to solve regarding removal.
`INFO:align_data.embeddings.pinecone.update_pinecone:Progress: 2.6% (300/11592)
ERROR:align_data.embeddings.pinecone.pinecone_db_handler:Got exception while retrying: (400)
Reason: Bad Request
HTTP response headers: HTTPHeaderDict({'Date': 'Mon, 05 May 2025 04:24:59 GMT', 'Content-Type': 'application/json', 'Content-Length': '109', 'Connection': 'keep-alive', 'x-pinecone-request-latency-ms': '32', 'x-pinecone-request-id': '5955277797286641668', 'x-envoy-upstream-service-time': '1', 'server': 'envoy'})
HTTP response body: {"code":3,"message":"Number of provided IDs: 1268 exceeds the maximum amount per request: 1000","details":[]}
`

LeMurphant:
I updated removals to batch them

mruwnik:
it's very possible that the --rebuild option used to exist, but got deleted from the code. And nobody remembered to update the README

LeMurphant:
> it's very possible that the --rebuild option used to exist, but got deleted from the code. And nobody remembered to update the README

@mruwnik this makes sense, I can remove it from the README. Do you know what would be the best way to actually clean up? Does cleanup.py makes sense?

mruwnik:
yup. You could trawl the git history for where it got removed, but I don't recommend that...

LeMurphant:
@jbeshir Do the last few commits address the issues you had mentioned [here](https://discord.com/channels/986243767424139294/1075908160214274059/1370807800590106745)?

align_data/embeddings/pinecone/update_pinecone.py
```diff
@@ -47,16 +51,51 @@ def update(self, custom_sources: List[str], force_update: bool = False):
...
+            logger.info("Processing %s items from %s", total_articles, custom_sources)
+
+            total_processed = 0
+            update_interval = self.batch_size
+
             for batch in self.batch_entries(articles_to_update):
                 self.save_batch(session, batch)
+                total_processed += len(batch)
+
+                if total_processed % update_interval == 0 or total_processed == total_articles:
```
jbeshir, line 65: Not sure under what circumstances this is expected to return false- if I follow right, looks like the first condition should pass on all batches other than the last one, since `batch_entries` breaks it into batches of length `batch_size`, and the second should pass on the last batch.

I'm not sure it works defensively in the case where that changes, because if a batch could be smaller than batch_size, then I think this would cause the first clause to fail for all subsequent runs so you'd only get the final update at the end after that.

Would it make sense to remove this check if the intended behaviour is to run after every batch?

jbeshir:
> @jbeshir Do the last few commits address the issues you had mentioned [here](https://discord.com/channels/986243767424139294/1075908160214274059/1370807800590106745)?

Makes it nicely split out and documents how to run it, which would be very convenient for me! I am still not sure about the update logging mechanism but that's not a major thing.

align_data/embeddings/pinecone/update_pinecone.py
LeMurphant: In the end I agree, printing the progression every time is probably equivalent, and even if it is not, it's fine. Fixed in https://github.com/StampyAI/alignment-research-dataset/pull/210/commits/c5befde57fef60d03b28c21b4023308bcfe4b163

align_data/embeddings/pinecone/pinecone_db_handler.py
```diff
@@ -230,22 +238,49 @@ def query_text(
...
+
+        Pinecone has a limit of 1000 IDs per delete request.
+        """
+        max_delete_size = 900 
+
+        if len(ids) <= max_delete_size:
+            self.index.delete(ids=ids, namespace=PINECONE_NAMESPACE)
+        else:
+            # If we exceed the limit, chunk the deletion
+            for i in range(0, len(ids), max_delete_size):
```
mruwnik, line 258: I'd just always go with this - no point in checking if the `ids` are smaller. If they are, the loop should handle them.
Have you quadrupely checked that this chunking works? I'm always afraid that this kind of stuff will have indexing errors

align_data/embeddings/pinecone/pinecone_db_handler.py
```diff
@@ -230,22 +238,49 @@ def query_text(
...
+        """Delete entries from Pinecone by their hash_ids.
+
+        Each hash_id may expand to multiple vector IDs. This method processes
+        article IDs in small chunks to avoid memory issues and handles the
+        expanded vector IDs to avoid exceeding Pinecone's 1000 ID limit per request.
+        """
+        article_chunk_size = 10
+
+        for i in range(0, len(ids), article_chunk_size):
+            article_chunk = ids[i : i + article_chunk_size]
```
mruwnik, line 273: you might consider doing somehing like
```
def chunk_items(items, chunk_size):
    for i in range(0, len(items), chunk_size):
         yield items[i : i + chunk_size]

(...)
for chunk in chunk_items(ids, article_chunk_size):
    <do stuff>
```
as that would make it easier to check that the tricky bit (chunking) works correctly

align_data/embeddings/pinecone/update_pinecone.py
```diff
@@ -47,16 +51,47 @@ def update(self, custom_sources: List[str], force_update: bool = False):
...
         """Update the Pinecone entries of specific articles based on their hash_ids."""
         with make_session() as session:
             articles_to_update = self._articles_by_id(session, hash_ids, force_update)
+            total_articles = articles_to_update.count()
+
+            logger.info("Processing %s items by ID", total_articles)
+
+            total_processed = 0
+
             for batch in self.batch_entries(articles_to_update):
```
mruwnik, line 83: you could add the printing to `batch_entires` behind a `log_progress` flag

align_data/embeddings/pinecone/pinecone_db_handler.py
```diff
@@ -230,22 +238,49 @@ def query_text(
...
+        return list(self.index.list(prefix=id_, namespace=PINECONE_NAMESPACE))
 
     @with_retry()
     def _del_items(self, ids):
-        self.index.delete(ids=ids, namespace=PINECONE_NAMESPACE)
+        """Delete vector IDs from Pinecone.
+
+        Pinecone has a limit of 1000 IDs per delete request.
+        """
+        max_delete_size = 900
```
mruwnik, line 252: why not use the full 1000?

align_data/embeddings/pinecone/update_pinecone.py
```diff
@@ -154,7 +191,9 @@ def _make_pinecone_entry(self, article: Article) -> PineconeEntry | None:
 
 
 class PineconeDeleter(PineconeAction):
-    batch_size = 100
+    # Using a smaller batch size to avoid exceeding Pinecone's 1000 ID limit
+    # Since each article can expand to multiple vector IDs when deleted
+    batch_size = 30
```
mruwnik, line 196: why not just let the method that deletes things handle the pinecone limit?

align_data/sources/greaterwrong/cleanup.py
```diff
@@ -0,0 +1,140 @@
...
+
+def check_post_meets_tag_requirements(post, tag_requirements):
+    if tag_requirements.get("bypass_tag_check", False):
+        return True
+
+    # Check if metadata contains tags
+    if not post.meta:
+        return False
+
+    meta = json.loads(post.meta) if isinstance(post.meta, str) else post.meta
```
mruwnik, line 28:
```
if not (meta := post.meta):
    return False

if isinstance(meta, str):
    meta = json.loads(meta)
````

align_data/embeddings/pinecone/pinecone_db_handler.py
LeMurphant: It was a safety buffer, but ended up not being necessary.

align_data/embeddings/pinecone/pinecone_db_handler.py
LeMurphant: I used your suggestion in https://github.com/StampyAI/alignment-research-dataset/pull/210/commits/2bf579c43ebbf91c45d5d1446c464dfe38764711 and added a bunch of tests, seems to work OK

align_data/embeddings/pinecone/update_pinecone.py
mruwnik: I was thinking more that the methods that are logging progress use `batch_entries`. So you could just have that method do the logging, rather than have it in multiple places

mruwnik (approved):
I assume it works properly?

align_data/embeddings/pinecone/update_pinecone.py
LeMurphant: Handled by https://github.com/StampyAI/alignment-research-dataset/pull/210/commits/b58246e75968ba3ee42470b5560a4f0749c9d566

LeMurphant:
> I assume it works properly?

Not fully tested yet, wanted make sure the logging was what you wanted. Going to test it now.

align_data/embeddings/pinecone/update_pinecone.py
LeMurphant: I added a helper function in https://github.com/StampyAI/alignment-research-dataset/pull/210/commits/8cce3872585f819109c34261c323ade5336f3eba since I would have to duplicate the logging code in the `batch_entries` functions of both `PineconeAction` and `PineconeAdder`. LMK if there's a better way to do it.

PR #236, opened 2026-05-30
Declare tenacity + document retry policy

nerualauren:
Two small housekeeping follow-ups to the retry-hardening work (#235).

## 1. Declare tenacity in requirements
`embedding_utils` imports tenacity but it was never declared in `requirements.txt`. Same latent-break class as the SDK-9 pinecone bug: works today only because tenacity is transitively present, would break a clean install. Pinned `>=8.0.0` (floor where all the imported names exist).

## 2. Add CLAUDE.md documenting the retry policy
Records the policy the #235 consolidation established, so the next contributor doesn't re-fragment it:
- simple stateless transient retry -> `common/retry.py:with_retry`
- Pinecone 429 ladder -> bespoke `with_pinecone_retry`
- stateful (DB rollback between attempts) -> bespoke loops

Plus a note to grep for existing variants before adding a new util (parsimony pass per change).

Needs a one-line `.gitignore` negation (`!CLAUDE.md`) since the repo ignores `*.md`, mirroring the existing `!requirements.txt`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


LeMurphant (approved):
Not sure about explicitly including instructions in claude.md to github, but I'll defer to you on that.

